PR #5, opened 2020-07-24
Added Footer Data

sumitd94:


pages/index.js
```diff
@@ -22,6 +22,14 @@ const Index = () => {
         />
       </Head>
       <HomePage />
+      <footer>
+        <p>
+          The contents of this website are deployed from this 
+          <a href="{https://github.com/Real-Dev-Squad/website-members}" target="_blank">
```
ankushdharkar, line 28: No css used?

sumitd94: Not yet, I first wanted to make sure this changes work and accordingly next step will be to add css.

ankushdharkar, line 28: Do you need `{ }` for html attribute?

sumitd94: To render the the href link in react we use { } . Will the link work without {} as well ? Let me try and check in my local

ankushdharkar: Not good from a PR submission standpoint. Finish adding the CSS please.

ankushdharkar: Without the { }, no dynamic content will render. But, our string is static anyway :)

sumitd94: Yep ok

sumitd94: Ok, I will add the css and re submit the PR

ankushdharkar:
Solves #9

ankushdharkar:
Closed in favour of #9

PR #19, opened 2020-07-26
Fetching and Showing members in the home page 

invalidtoken:
1. Fetching and showing members on the homepage 
2. Linking each member to their corresponding `/members/[id]` link.

Screenshot 2020-07-26 at 9 44 04 PM


components/members-list/index.js
```diff
@@ -0,0 +1,37 @@
...
+import MemberListItem from 'components/member-list-item';
+import Loader from 'components/loader';
+
+import useFetch from 'custom-hooks/useFetch';
+import classNames from './member-list.module.scss';
+
+const URL = 'https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main/ids/mapping.json';
+
+const MembersList = () => {
+  let { loading, error, data } = useFetch(URL);
```
shihabus, line 11: at least first set can be server fetched, and later on the client fetched

custom-hooks/useFetch.js
```diff
@@ -0,0 +1,57 @@
+import { useState, useEffect } from 'react';
```
shihabus, line 1: a best use case for `useReducer`

invalidtoken: Yes, `useReducer` is definitely the best option, but I didn't want to make the code complex for beginners to read. Already hooks are a little hard to understand and using `useReducer` will make the code less beginner-friendly.

components/members-list/index.js
invalidtoken: We are using Static-Build, we cannot use any server fetching features of Next.js.

shihabus: I forget, we not server-rendered.

custom-hooks/useFetch.js
shihabus: If you handle multiple states, useReducers are cleaner. The code is easier to reason about.

pages/_document.js
```diff
@@ -10,11 +10,10 @@ class MyDocument extends Document {
     return (
       <Html>
         <Head>
-          <meta charset="UTF-8" />
+          <meta charSet="UTF-8" />
```
ankushdharkar, line 13: Why is this camelcased?
I thought previous was fine: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta

custom-hooks/useFetch.js
invalidtoken: Valid argument @shihabus. Adding useReducer instead of multiple state changes.

helper-functions/urls.js
```diff
@@ -0,0 +1,15 @@
...
+const getImgURL = (rdsId) =>
+  `https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main/members/${rdsId}/img.png`;
+
+/**
+ *
+ * @param {string} rdsId
+ */
+const getDataURL = (rdsId) =>
+  `https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main/members/${rdsId}/data.json`;
+
```
ankushdharkar, line 14: Let's move `https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main/members` to `baseURL` constant

pages/_document.js
invalidtoken: Its the difference b/w HTML and react. React has camel-cased everything.

helper-functions/urls.js
invalidtoken: Moved to baseURL constant.

pages/_document.js
ankushdharkar: Thought so. Ok, makes sense. Thanks

custom-hooks/useFetch.js
ankushdharkar: @invalidtoken For now, let's submit what you've done. Make an enhancement-tagged issue and point to these points please. We can pick it up in future.

ankushdharkar: Added it: https://github.com/Real-Dev-Squad/website-members/issues/20

helper-functions/urls.js
```diff
@@ -1,15 +1,15 @@
...
 /**
  *
  * @param {string} rdsId
  */
-const getDataURL = (rdsId) =>
-  `https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main/members/${rdsId}/data.json`;
+const getDataURL = (rdsId) => `${baseURL}/members/${rdsId}/data.json`;
 
-export { getImgURL, getDataURL };
+export { getImgURL, getDataURL, baseURL };
```
ankushdharkar, line 15: Why are we exporting it?

invalidtoken: Instead of making `https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main/members` this the baseURL I made `https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main` this the base URL so in future we can just add things to this URL like `${baseURL}/ids/mapping.json` etc.

ankushdharkar: Haan, but we should not export it from a utils file. We should put this in constants file and import (can be done in future)

Utils files is only functions, not properties

ankushdharkar: Yes, but we should not export it from a utils file. We should put this in constants file and import (can be done in future)

Utils files is only functions, not properties

invalidtoken: Make sense, baseURL removed from export.

PR #21, opened 2020-07-26
member list hover minor css update for members

Harshi7016:
image

I have made minor css updates

### DIFF
--- a/components/member-list-item/member-list-item.module.scss
+++ b/components/member-list-item/member-list-item.module.scss
@@ -33,3 +33,8 @@
     width: 28px;
   }
 }
+
+.container:hover {
+  box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
+  transform: translateY(-2px);
+}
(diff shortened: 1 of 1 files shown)


Harshi7016:
I observed the last card in the page gets css issue
image

Harshi7016:
I didn't get the time to check for this issue yet
I am working on it

components/member-list-item/member-list-item.module.scss
```diff
@@ -33,3 +33,8 @@
     width: 28px;
   }
 }
+
+.container:hover {
+  box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
+  transform: translateY(-10px);
+}
```
ankushdharkar, line 40: Hovering indicates we can do something with the card, like clicking it our picking it. Is that what we expect to allow the user?

Harshi7016: Yeah, I mean after going through the basic structure, I observed on clicking the image we would be going to the details page.
Either that or I can just add box shadow to the card and remove the transition on hovering....
Open to suggestion

ankushdharkar: So, we are doing something on clicking? Got it. Thanks for explaining @Harshi7016 :)

PR #36, opened 2020-07-31
remove extra metatags

ashishmangla0:


pages/_document.js
```diff
@@ -10,8 +10,6 @@ class MyDocument extends Document {
     return (
       <Html>
         <Head>
-          <meta charSet="UTF-8" />
-          <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
sumitd94, line 14: Any reason why are we removing this ?

sumitd94 (changes requested):

pages/_document.js
ashishmangla0: bcz, this are the duplicate tags. next have this tags already

PR #37, opened 2020-07-31
CSS fixes

tejasberi:
- Making the layout enhancements
- Adding missing prop-types
- Adding the CSS transition on tile

### DIFF
--- a/components/member-list-item/index.js
+++ b/components/member-list-item/index.js
@@ -26,7 +26,7 @@ const PreviewMember = (props) => {
         </a>
       </Link>
 
-      <h2 className={classNames.name}>
+      <h2 className={classNames.nameOfPerson}>
         {data ? `${data['first_name']} ${data['last_name']}` : rdsId}
       </h2>
       {data && (
--- a/components/member-list-item/member-list-item.module.scss
+++ b/components/member-list-item/member-list-item.module.scss
@@ -5,11 +5,17 @@
   margin: 20px;
   border: 1px solid #f4f4f4;
   padding: 20px 10px;
+  background: white;
 
   //addition of box-shadow and border radius to member card
   box-shadow: 10px 10px 20px -10px rgba(0, 0, 0, 0.22);
   border-radius: 5%;
 
+  &:hover {
+    box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
+    transform: translateY(-10px);
+    transition-duration: 0.5s;
+  }
 }
 
 .imgContainer {
@@ -20,10 +26,9 @@
   background-position: center;
   background-repeat: no-repeat;
   background-size: cover;
-}
-//transition of image for better user experiance
-.imgContainer:hover {
-  transform: translateY(-5px);
+  -webkit-box-shadow: 0px 0px 15px 1px rgba(0, 0, 0, 0.75);
+  -moz-box-shadow: 0px 0px 15px 1px rgba(0, 0, 0, 0.75);
+  box-shadow: 0px 0px 10px 1px rgba(0, 0, 0, 0.75);
 }
 
 .iconsContainer {
@@ -43,4 +48,8 @@
   img:hover {
     transform: scale(1.1);
   }
+}
+
+.nameOfPerson {
+  margin: 16px 0;
 }
--- a/components/member-profile/index.js
+++ b/components/member-profile/index.js
@@ -1,5 +1,7 @@
-import React from 'react';
 import { getDataURL, getImgURL } from 'helper-functions/urls';
+
+import PropTypes from 'prop-types';
+import React from 'react';
 import classNames from './member-profile.module.scss';
 import useFetch from 'custom-hooks/useFetch';
 
@@ -8,12 +10,20 @@ const Profile = (props) => {
   const { data } = useFetch(getDataURL(props.id));
   return (
     <>
-      <img src={imgLink} className={classNames.profilePic} />
+      <img src={imgLink} className={classNames.profilePic} alt="profile" />
       <div className={classNames.memberName}>
         {data ? `${data['first_name']} ${data['last_name']} ` : `${props.id}`}
       </div>
     </>
   );
+};
+
+Profile.propTypes = {
+  id: PropTypes.string
+};
+
+Profile.defaultProps = {
+  id: ''
 };
 
 export default Profile;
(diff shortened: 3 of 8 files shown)


components/member-profile/index.js
```diff
@@ -8,12 +10,20 @@ const Profile = (props) => {
   const { data } = useFetch(getDataURL(props.id));
   return (
     <>
-      <img src={imgLink} className={classNames.profilePic} />
+      <img src={imgLink} className={classNames.profilePic} alt="profile" />
```
shihabus, line 13: alt="profile image"

components/member-list-item/member-list-item.module.scss
```diff
@@ -5,11 +5,17 @@
...
   border: 1px solid #f4f4f4;
   padding: 20px 10px;
+  background: white;
 
   //addition of box-shadow and border radius to member card
   box-shadow: 10px 10px 20px -10px rgba(0, 0, 0, 0.22);
   border-radius: 5%;
 
+  &:hover {
+    box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
```
shihabus, line 15: good to have `-webkit-box-shadow:` and `-moz-box-shadow:`

components/member-list-item/index.js
```diff
@@ -26,7 +26,7 @@ const PreviewMember = (props) => {
         </a>
       </Link>
 
-      <h2 className={classNames.name}>
+      <h2 className={classNames.nameOfPerson}>
         {data ? `${data['first_name']} ${data['last_name']}` : rdsId}
```
shihabus, line 30: default first name and last name like `${data['first_name']}||'--` and `${data['last_name']}||'--`

components/member-profile/index.js
```diff
@@ -8,12 +10,20 @@ const Profile = (props) => {
   const { data } = useFetch(getDataURL(props.id));
```
shihabus, line 10: NIT: destructure id like `const {id}=props`

styles/layout-style.module.scss
```diff
@@ -7,4 +7,5 @@
 
 .main {
   margin: 20px;
+  font-family: monospace;
```
shihabus, line 10: NIT: use the same font throughout, unless required so.

components/member-list-item/index.js
Harshi7016: @tejasberi I have added padding 10px 10px for classname as name  itself

components/member-list-item/member-list-item.module.scss
Harshi7016: good addition (y)

ankushdharkar: @Shihab-Duzno @Harshi7016 Always check: https://caniuse.com/#search=box-shadow before adding prefixes. Usually build tools like autoprefixer do a good job with this, but right now, we have to be careful of unnecessarily increasing the size of our CSS for very little RoI
For box-shadow, I don't see a need for prefixing

Harshi7016: this is useful (y)

components/member-profile/index.js
tejasberi: good suggestion. 
Due to the linting errors set on the project, its giving error whein i give alt name as "profile image"
`error  Redundant alt attribute. Screen-readers already announce `img` tags as an image. You don’t need to use the words `image`, `photo,` or `picture` (or any specified custom words) in the alt prop`

tejasberi: Thats the reason the alt name is profile and not profile image

components/member-list-item/member-list-item.module.scss
tejasberi: Agreed with @ankushdharkar 
Auto prefixer should take care of -webkit-box-shadow: and -moz-box-shadow

components/member-profile/index.js
tejasberi: @Shihab-Duzno request you to provide comments on code that i have changed. This line does not seem to be related to my changes.

styles/layout-style.module.scss
tejasberi: .main class i a wrapper for whole web-app; which means that its a single source of font-family.

components/member-list-item/index.js
tejasberi: @Shihab-Duzno your comment is not related to my code change in this PR

tejasberi: @harshi7016 we discussed about the class and agreed to use a unique name

styles/layout-style.module.scss
shihabus: Then better let it be in global styles file.

components/member-list-item/member-list-item.module.scss
shihabus: Agreed

styles/layout-style.module.scss
tejasberi: .main itself is a wrapper component. The intension is to have the parent styles in the parent component. 
Also, request you to be be polite in ur comments and we are trying to help each other and not teach each other. Holding a simple PR for NIT comments is not suggested. Hope u get the polite message.

shihabus: Sorry if I sounded rude. I am not doing this to be sounded rude either. Since this is a learning process, I am too learning along with you. No offense intended.

tejasberi: @ankushdharkar request for ur review

components/member-list-item/member-list-item.module.scss
ankushdharkar: Even if we have to put it ourselves, I am encouraging you all to check and put only the ones that are really needed. Don't put 4 versions just because you think it would cover the cases. The browsers themselves may not be implementing that variant anymore. Your code has no impact then.
@Shihab-Duzno @tejasberi @Harshi7016

styles/layout-style.module.scss
ankushdharkar: I lack context here. @tejasberi Could you explain to us please why we are using `monospace` font separately from what's coming in from the site?

tejasberi: i have tried to play with a new font here. seem to be looking good. suggest u to gave a look. if that does not suit the site style guide, i will revert it.

ankushdharkar: Got it. Thanks

PR #38, opened 2020-07-31
Add Layout Component

### LINKED ISSUE #32: Page title missing
We need to title to all pages.

whyDontI:
# This fixes #32 
## Created

- Layout component

## Modified

- `pages/index.js`
- `pages/members/[id]/index.js`

## Fixed linting errors

- `components/member-profile/index.js`

### DIFF
--- a/components/layout/index.js
+++ b/components/layout/index.js
@@ -1,0 +1,30 @@
+import Head from 'next/head';
+import PropTypes from 'prop-types';
+
+const Layout = (props) => {
+  const { children, title, description } = props;
+  return (
+    <div>
+      <Head>
+        <title>{title}</title>
+        <meta name="description" content={description} />
+      </Head>
+      <div className="container">
+        <div>{children}</div>
+      </div>
+    </div>
+  );
+};
+
+Layout.propTypes = {
+  children: PropTypes.any,
+  title: PropTypes.string,
+  description: PropTypes.string
+};
+
+Layout.defaultProps = {
+  title: 'Real Dev Squad',
+  description: ''
+};
+
+export default Layout;
--- a/components/member-profile/index.js
+++ b/components/member-profile/index.js
@@ -2,18 +2,28 @@ import React from 'react';
 import { getDataURL, getImgURL } from 'helper-functions/urls';
 import classNames from './member-profile.module.scss';
 import useFetch from 'custom-hooks/useFetch';
+import PropTypes from 'prop-types';
 
 const Profile = (props) => {
-  const imgLink = getImgURL(props.id);
-  const { data } = useFetch(getDataURL(props.id));
+  const { id } = props;
+  const imgLink = getImgURL(id);
+  const { data } = useFetch(getDataURL(id));
   return (
     <>
-      <img src={imgLink} className={classNames.profilePic} />
+      <img src={imgLink} alt="" className={classNames.profilePic} />
       <div className={classNames.memberName}>
-        {data ? `${data['first_name']} ${data['last_name']} ` : `${props.id}`}
+        {data ? `${data['first_name']} ${data['last_name']} ` : `${id}`}
       </div>
     </>
   );
+};
+
+Profile.propTypes = {
+  id: PropTypes.string
+};
+
+Profile.defaultProps = {
+  id: ''
 };
 
 export default Profile;
--- a/components/pages/index/index.js
+++ b/components/pages/index/index.js
@@ -1,14 +1,10 @@
 import React from 'react';
-import Head from 'next/head';
 import MemberList from 'components/members-list';
 import classNames from './home-page.module.scss';
 
 const HomePage = () => {
   return (
     <div className={classNames.container}>
-      <Head>
-        <title>Members | Real Dev Squad</title>
-      </Head>
       <img className={classNames.img} src="/images/Real-Dev-Squad@1x.png" alt="real-dev squad" />
       <h1 className={classNames.heading}>Real Dev Squad Members</h1>
       <MemberList />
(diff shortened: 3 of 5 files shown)


components/layout/index.js
```diff
@@ -0,0 +1,30 @@
...
+      </Head>
+      <div className="container">
+        <div>{children}</div>
+      </div>
+    </div>
+  );
+};
+
+Layout.propTypes = {
+  children: PropTypes.any,
```
shihabus, line 20: `children: PropTypes.node.isRequired,`

components/member-profile/index.js
```diff
@@ -2,18 +2,28 @@ import React from 'react';
 import { getDataURL, getImgURL } from 'helper-functions/urls';
 import classNames from './member-profile.module.scss';
 import useFetch from 'custom-hooks/useFetch';
+import PropTypes from 'prop-types';
```
shihabus, line 5: NIT: lib imports on top, followed by components and local

components/member-profile/index.js
```diff
@@ -2,18 +2,28 @@ import React from 'react';
...
 const Profile = (props) => {
-  const imgLink = getImgURL(props.id);
-  const { data } = useFetch(getDataURL(props.id));
+  const { id } = props;
+  const imgLink = getImgURL(id);
+  const { data } = useFetch(getDataURL(id));
   return (
     <>
-      <img src={imgLink} className={classNames.profilePic} />
+      <img src={imgLink} alt="" className={classNames.profilePic} />
```
shihabus, line 13: don't leave alt empty

components/member-profile/index.js
```diff
@@ -2,18 +2,28 @@ import React from 'react';
...
       <div className={classNames.memberName}>
-        {data ? `${data['first_name']} ${data['last_name']} ` : `${props.id}`}
+        {data ? `${data['first_name']} ${data['last_name']} ` : `${id}`}
```
shihabus, line 15: what if data don't have `first_name` or `last_name`.

{data ? `${data['first_name']||'--'} ${data['last_name']||'--'} ` : `${id}`}

components/pages/index/index.js
```diff
@@ -1,14 +1,10 @@
...
 import MemberList from 'components/members-list';
 import classNames from './home-page.module.scss';
 
 const HomePage = () => {
   return (
     <div className={classNames.container}>
-      <Head>
-        <title>Members | Real Dev Squad</title>
-      </Head>
       <img className={classNames.img} src="/images/Real-Dev-Squad@1x.png" alt="real-dev squad" />
```
shihabus, line 8: NIT: this should a clickable image which takes the user to the home-page of RDS

pages/members/[id]/index.js
```diff
@@ -1,10 +1,14 @@
 import React from 'react';
 import { useRouter } from 'next/router';
 import Profile from 'components/member-profile';
-
+import Layout from 'components/layout';
 const MemberProfile = () => {
   let { query } = useRouter();
-  return <Profile id={query.id} />;
+  return (
+    <Layout title={`${query.id} | Member Real Dev Squad`}>
```
Comment (shihabus) [inline pages/members/[id]/index.js:8]: why id? It can the first name and `| RDS`.

components/layout/index.js
```diff
@@ -0,0 +1,30 @@
...
+  const { children, title, description } = props;
+  return (
+    <div>
+      <Head>
+        <title>{title}</title>
+        <meta name="description" content={description} />
+      </Head>
+      <div className="container">
+        <div>{children}</div>
+      </div>
```
shihabus, line 14: If we have a layout, can't we have the `Footer` here? We can give the `.container` a min-height of `95vh`, so that footer is always at the bottom. Coordinate with @manjots1607

pages/members/[id]/index.js
```diff
@@ -1,10 +1,14 @@
 import React from 'react';
```
Comment (shihabus) [inline pages/members/[id]/index.js:1]: I guess react is available by default in Next. please confirm.

pages/members/[id]/index.js
```diff
@@ -1,10 +1,14 @@
...
   let { query } = useRouter();
-  return <Profile id={query.id} />;
+  return (
+    <Layout title={`${query.id} | Member Real Dev Squad`}>
+      <Profile id={query.id} />
+    </Layout>
+  );
 };
 
 export default MemberProfile;
```
Comment (shihabus) [inline pages/members/[id]/index.js:14]: Wrap in `React.memo()`

pages/members/[id]/index.js
```diff
@@ -1,10 +1,14 @@
 import React from 'react';
 import { useRouter } from 'next/router';
 import Profile from 'components/member-profile';
-
+import Layout from 'components/layout';
 const MemberProfile = () => {
```
Comment (shihabus) [inline pages/members/[id]/index.js:5]: I guess arrow function will be always rebuilt, try using normal function. @ankushdharkar any comments.

pages/members/[id]/index.js
Comment (whyDontI) [reply pages/members/[id]/index.js:8]: Then we need to request User data here just for first_name.
And requesting data from here and passing the data as prop won't be good.
because whille directly visiting `rds.com/members/shihabus` the component will not have any data.
Should we request the data twice?

components/layout/index.js
whyDontI: Better I'll talk to him.

pages/members/[id]/index.js
Comment (shihabus) [reply pages/members/[id]/index.js:8]: Can't you pass it as query along from the main page?

pages/members/[id]/index.js
Comment (whyDontI) [reply pages/members/[id]/index.js:5]: We are using Arrow functions in other files too,
And I did not work on the function, I have only added the layout component.
If it is necessary we can always make new issue and work on it.

PR #44, opened 2020-08-03
Feature/page title

### LINKED ISSUE #32: Page title missing

whyDontI:
Made suggested changes from last review - https://github.com/Real-Dev-Squad/website-members/pull/38

Moved footer from _app.js to Layout component 

This fixes #32


pages/index.js
```diff
@@ -1,10 +1,11 @@
 import HomePage from '../components/pages/index';
+import Layout from '../components/layout/index';
```
ankushdharkar, line 2: Why not just `/layout`?

pages/members/[id]/index.js
```diff
@@ -1,10 +1,19 @@
...
+import Layout from 'components/layout';
 const MemberProfile = () => {
   let { query } = useRouter();
-  return <Profile id={query.id} />;
+  const { first_name, last_name } = query;
+
+  return (
+    <Layout
+      title={`${
+        query.first_name && query.last_name ? `${first_name} ${last_name}` : query.id
```
Comment (ankushdharkar) [inline pages/members/[id]/index.js:12]: Could we have this a bit cleaner please?

Something like
```js
const { first_name: firstName, last_name: lastName, id: queryID } = query;
const hasBothNames = firstName && lastName;

const titleStr = hasBothNames ? `${firstName} ${lastName}` : queryID
```

components/member-profile/index.js
```diff
@@ -1,18 +1,18 @@
-import { getDataURL, getImgURL } from 'helper-functions/urls';
-
-import PropTypes from 'prop-types';
 import React from 'react';
-import classNames from './member-profile.module.scss';
+import PropTypes from 'prop-types';
```
ankushdharkar, line 2: Was this used?

components/member-profile/index.js
```diff
@@ -1,18 +1,18 @@
...
-  const { data } = useFetch(getDataURL(props.id));
+  const { id } = props;
+  const imgLink = getImgURL(id);
+  const { data } = useFetch(getDataURL(id));
   return (
     <>
       <img src={imgLink} className={classNames.profilePic} alt="profile" />
       <div className={classNames.memberName}>
-        {data ? `${data['first_name']} ${data['last_name']} ` : `${props.id}`}
+        {data ? `${data['first_name'] || '--'} ${data['last_name'] || '--'} ` : `${id}`}
```
ankushdharkar, line 15: Let's move this into a single variable above the `return` please

components/member-list-item/index.js
```diff
@@ -19,25 +19,35 @@ const PreviewMember = (props) => {
...
-    </Link>
+      <h2 className={classNames.nameOfPerson}>
+        {data ? `${data['first_name']} ${data['last_name']}` : rdsId}
+      </h2>
+      {data && (
+        <div className={classNames.iconsContainer}>
+          {data['twitter_id'] && <SocialMediaIcon id={data.twitter_id} type="twitter" />}
+          {data['github_id'] && <SocialMediaIcon id={data.github_id} type="github" />}
+          {data['linkedin_id'] && <SocialMediaIcon id={data.linkedin_id} type="linkedIn" />}
+          {data['instagram_id'] && <SocialMediaIcon id={data.instagram_id} type="instagram" />}
```
ankushdharkar, line 47: NIT: Let's move the "complex" variable conditionals above the `return`

e.g
```js
const { twitter_id: twitterID } = data;

return {
 { twitterID && <SocialMediaIcon id={twitterID} /> } 
}
```

pages/members/[id]/index.js
Comment (whyDontI) [reply pages/members/[id]/index.js:12]: Spreading the query object is ok, But Sahil(@invalidtoken ) has used `first_name` and `last_name` in snake_case, so I just kept it that way.
Should I convert all the `first_name` and `last_name`s to camelCase?, If yes then we also won't need to rename object keys.

components/member-profile/index.js
whyDontI: Yes it is used on the line number 21

components/member-list-item/index.js
whyDontI: Can we have another issue/PR for this as this is not related to Page Title?

ankushdharkar: I can see that this code was touched in this PR @whyDontI

components/member-profile/index.js
ankushdharkar: Oh. My bad. How was this used before importing earlier 🤦

pages/members/[id]/index.js
Comment (ankushdharkar) [reply pages/members/[id]/index.js:12]: Backend replies in snake_case, in JS, we adapt it to camelCase (this helps in quickly assessing if the data is coming from an API, or our JS world)

Will add this in the code docs

components/member-list-item/index.js
whyDontI: I think I have just moved it around I guess.
@tejasberi has worked on it.
Screenshot from 2020-08-12 11-54-57

tejasberi: whats the issue here?

components/member-profile/index.js
whyDontI: Not sure 😁

Issue #50, opened 2020-08-08
Render member details on SSR during request side


ankushdharkar:
Currently, we fetch all the members on client side via N fetch requests. This is bad for the UX, and we want to create these in the page during request time (or available in cache)

So, during Request time, fetch the N members on the server and then serve the HTML page

PR #52, opened 2020-08-10
Moved Members Details page from SSG to SSR

sumitd94:
This fixes the issue https://github.com/Real-Dev-Squad/website-members/issues/50


pages/members/[id]/index.js
```diff
@@ -1,10 +1,28 @@
...
-const MemberProfile = () => {
-  let { query } = useRouter();
-  return <Profile id={query.id} />;
+const MemberProfile = ({ imageLink, data }) => {
+  return <Profile imageLink={imageLink} membersData={data} />;
+};
+
+export async function getServerSideProps(context) {
+  const imgLink = getImgURL(context.params.id);
+  const jsonUrl = getDataURL(context.params.id);
```
Comment (invalidtoken) [inline pages/members/[id]/index.js:13]: You can use Object Destructuring here.

pages/members/[id]/index.js
```diff
@@ -1,10 +1,28 @@
...
+
+  return { props: { imageLink: imgLink, data: returnData } };
+}
+
+MemberProfile.propTypes = {
+  imageLink: PropTypes.string
+};
+
+MemberProfile.propTypes = {
+  data: PropTypes.object
```
Comment (invalidtoken) [inline pages/members/[id]/index.js:25]: Why are here two propTypes? Both `imageLink` and `data` can be defined inside one. 
Also, defaultProps are missing.

components/member-profile/index.js
```diff
@@ -1,29 +1,29 @@
...
 import React from 'react';
 import classNames from './member-profile.module.scss';
-import useFetch from 'custom-hooks/useFetch';
 
 const Profile = (props) => {
-  const imgLink = getImgURL(props.id);
-  const { data } = useFetch(getDataURL(props.id));
+  const memberName = `${props.membersData.first_name || '--'} ${
+    props.membersData.last_name || '--'
+  }`;
```
invalidtoken, line 8: Is there any specific reason to do this?
Also, you can use destructure props. 
```
const { memberdData, imageLink } = props;
```

components/member-profile/index.js
```diff
@@ -1,29 +1,29 @@
...
 };
 
 Profile.propTypes = {
-  id: PropTypes.string
+  imageLink: PropTypes.string
+};
+
+Profile.propTypes = {
+  membersData: PropTypes.object
 };
```
invalidtoken, line 23: Why are here two propTypes? Both `membersData` and `imageLink` can be defined inside one propTypes object.

components/member-profile/index.js
```diff
@@ -1,29 +1,29 @@
...
 
 Profile.defaultProps = {
-  id: ''
+  imageLink: ''
```
invalidtoken, line 26: defaultProp for `membersData` is missing. Either make `membersData` as required or provide a defaultProp.

invalidtoken:
Also, one way to handle error for an invalid userName is like this - 

```
const MemberProfile = ({ imageLink, data, errorMsg }) => {
  if (errorMsg) {
    return <ErrorPage />;
  }
  return <Profile imageLink={imageLink} membersData={data} />;
};

export async function getServerSideProps(context) {
  const imgLink = getImgURL(context.params.id);
  const jsonUrl = getDataURL(context.params.id);
  try {
    const res = await fetch(jsonUrl);
    if (res.status !== 200) {
      throw new Error('Invalid Username');
    }

    const returnData = await res.json();
    return { props: { imageLink: imgLink, data: returnData } };
  } catch (error) {
    return { props: { errorMsg: error.message } };
  }
}
```

components/member-profile/index.js
sumitd94: Yes ok I will destruct the props.
The reason why I am displaying in this way is because shihab once said that what if no first_name or last_name exists.

shihabus: Yes, it is always good to have fallback values. But here you can do something like,

```
const { membersData:{first_name, last_name} }=props; // de-structure all props on top
const memberName=`${first_name||''} ${last_name||''}`;  // to avoid -- when either one is missing
```

while accessing `{memberName.trim()||'--'}`

_On a general note_
accessing values like `props. membersData.first_name` is easy to break. So better to use it like `props. membersData?.first_name`

components/member-profile/index.js
shihabus: if you know the prop structure, try to be specific in `propTypes`.

For example,

```
Profile.propTypes={
  membersData:PropTypes. shape({
        first_name: PropTypes.string,
        last_name: PropTypes.string
  })
}
```

If not required, default it too.

pages/members/[id]/index.js
```diff
@@ -1,10 +1,45 @@
+import { getDataURL, getImgURL } from 'helper-functions/urls';
+import PropTypes from 'prop-types';
+import fetch from 'cross-fetch';
 import React from 'react';
-import { useRouter } from 'next/router';
 import Profile from 'components/member-profile';
+import NotFound from 'components/404page';
```
Comment (ankushdharkar) [inline pages/members/[id]/index.js:6]: NIT: `not-found-page`, than calling it `404page`

components/member-profile/index.js
ankushdharkar: @sumitd94 Please follow @shihabus's recommendation

components/member-profile/index.js
```diff
@@ -1,29 +1,25 @@
-import { getDataURL, getImgURL } from 'helper-functions/urls';
-
 import PropTypes from 'prop-types';
 import React from 'react';
```
shihabus, line 2: NIT: React is available by default in Next

components/member-profile/index.js
```diff
@@ -1,29 +1,25 @@
...
 import classNames from './member-profile.module.scss';
-import useFetch from 'custom-hooks/useFetch';
 
 const Profile = (props) => {
-  const imgLink = getImgURL(props.id);
-  const { data } = useFetch(getDataURL(props.id));
+  const { membersData, imageLink } = props;
+  const memberName = `${membersData.first_name || '--'} ${membersData.last_name || '--'}`;
```
shihabus, line 7: Great if you refer my old comment. 

If the memberData is missing either `last_name` or `first_name` she/he will see `--`, which is not good.

pages/members/[id]/index.js
```diff
@@ -1,10 +1,45 @@
...
+const MemberProfile = ({ imageLink, data, errorMessage }) => {
+  if (errorMessage) {
+    return <NotFound />;
+  }
+  return <Profile imageLink={imageLink} membersData={data} />;
+};
+
+export async function getServerSideProps(context) {
+  const { params } = context;
+  const imgLink = getImgURL(params.id);
```
Comment (shihabus) [inline pages/members/[id]/index.js:17]: NIT: if you name the variable as `imageLink`, line 27 can be just `return { props:{ imageLink, data } }`.

Rename `returnData` to `data`.

components/member-profile/index.js
sumitd94: Got It, Thank you

components/member-profile/index.js
sumitd94: Did not understand this part.. could you please explain more on this.

pages/members/[id]/index.js
Comment (sumitd94) [reply pages/members/[id]/index.js]: @ankushdharkar Done, Please check

components/member-profile/index.js
sumitd94: @Shihab-Duzno Does this mean we are defining the properties of object (in this case membersData) more clearly ?

shihabus: @sumitd94 yes. It is to define the expected structure and type of data. If the data type is something different we can get warnings too. Some kind of type checking.

When default the value also, keep the same structure.

sumitd94: Defaulting like this : 

Profile.defaultProps = {
  imageLink: '',
  membersData: PropTypes.shape({
    first_name: 'Sumit',
    last_name: 'Sumit'
  })
};

Is this the right way ? @Shihab-Duzno

Shihab-Duzno:
```
Profile.defaultProps = {
  imageLink: '',
  membersData:{
    first_name: '',
    last_name: ''
  }
};
```

sumitd94: Ohh I see, Got it
Thank you :+1:

invalidtoken:
This looks real good @sumitd94. Good Job. 
@ankushdharkar @shihabus Shouldn't we add the link to the home page? As a user, if something is not found I look for a link to the current place.

Screenshot 2020-08-11 at 10 07 02 PM

pages/members/[id]/index.js
```diff
@@ -1,10 +1,46 @@
...
-  return <Profile id={query.id} />;
+const MemberProfile = ({ imageLink, data, errorMessage }) => {
+  if (errorMessage) {
+    return <NotFound errorMsg={errorMessage} />;
+  }
+  return <Profile imageLink={imageLink} membersData={data} />;
+};
+
+export async function getServerSideProps(context) {
+  const { params } = context;
```
Comment (shihabus) [inline pages/members/[id]/index.js:15]: NIT: `const { params: {id} } = context`

pages/members/[id]/index.js
```diff
@@ -1,10 +1,46 @@
-import React from 'react';
-import { useRouter } from 'next/router';
+import { getDataURL, getImgURL } from 'helper-functions/urls';
+import PropTypes from 'prop-types';
+import fetch from 'cross-fetch';
```
Comment (shihabus) [inline pages/members/[id]/index.js:3]: NIT: `fetch` is available by default.

sumitd94:
> This looks real good @sumitd94. Good Job.
> @ankushdharkar @shihabus Shouldn't we add the link to the home page? As a user, if something is not found I look for a link to the current place.
> Other then this I think all the changes are absolutely fine, again good job @sumitd94.
> 
> Screenshot 2020-08-11 at 10 07 02 PM

@invalidtoken I had the same thought of making the "members" word a href link where we will redirect the user to main home page .. but I was not able to figure it out how can we add a href link to the message. (so basically the way we are setting the message, I was not sure how to insert a href tag over there)
I was figuring out a way for this

pages/members/[id]/index.js
Comment (sumitd94) [reply pages/members/[id]/index.js:3]: If I am not importing fetch here, I was getting an error saying "Fetch is Not defined"
And that is why I imported fetch :)

components/member-profile/index.js
```diff
@@ -1,29 +1,34 @@
...
+  const memberName = `${first_name || ''} ${last_name || ''}`;
   return (
     <>
-      <img src={imgLink} className={classNames.profilePic} alt="profile" />
-      <div className={classNames.memberName}>
-        {data ? `${data['first_name']} ${data['last_name']} ` : `${props.id}`}
-      </div>
+      <img src={imageLink} className={classNames.profilePic} alt="profile" />
+      <div className={classNames.memberName}>{memberName.trim() || '--'}</div>
     </>
```
invalidtoken, line 14: Add `Head` with a title here, otherwise next.js shows the route on the title. 
To see example look [here](https://nextjs.org/docs/api-reference/next/head).
```
<Head>
  <title>{memberName} Profile<title>
</Head>
```

components/not-found-page/index.js
```diff
@@ -1,13 +1,25 @@
+import PropTypes from 'prop-types';
 import classNames from './not-found-page.module.scss';
 
-export default function index() {
+const Index = ({ errorMsg }) => {
   return (
     <div className={classNames.fullPageContainer}>
```
invalidtoken, line 6: Please add`Head` here as well. 
```
<Head>
  <title>Not Found</title>
</Head>
```

invalidtoken:
> > This looks real good @sumitd94. Good Job.
> > @ankushdharkar @shihabus Shouldn't we add the link to the home page? As a user, if something is not found I look for a link to the current place.
> > Other then this I think all the changes are absolutely fine, again good job @sumitd94.
> > Screenshot 2020-08-11 at 10 07 02 PM
> 
> @invalidtoken I had the same thought of making the "members" word a href link where we will redirect the user to main home page .. but I was not able to figure it out how can we add a href link to the message. (so basically the way we are setting the message, I was not sure how to insert a href tag over there)
> I was figuring out a way for this

@sumitd94 yes I think we can skip this for not lets finish and merge this PR as soon as possible. 
The final call is upon @ankushdharkar.

components/member-profile/index.js
sumitd94: @whyDontI has his PR ready with these changes

components/not-found-page/index.js
sumitd94: @whyDontI has his PR ready for this changes

invalidtoken: Okay cool.

pages/members/[id]/index.js
Comment (shihabus) [reply pages/members/[id]/index.js:3]: > You still need to import `isomorphic-fetch` or `isomorphic-unfetch` if you're using fetch() on the server.

My bad.

Issue #57, opened 2020-08-13
Introduce Pre-commit/push hooks


whyDontI:
# Add pre-commit hook to
- lint
- check branch names 

# Add pre-push hook to
- pull code from upstream / origin

whyDontI:
Discussion is pending about branch name conventions

PR #62, opened 2020-08-16
Introduce git-hooks

whyDontI:
Solves #57 
Added
- Pre Commit hook to check for
    - linting
    - branch name
- Pre Push hook to pull from upstream/origin before push

Code is commented because I am not sure what should be the name of remote (it will depend on the developer)

### DIFF
--- a/util/generic-functions.util.js
+++ b/util/generic-functions.util.js
@@ -1,0 +1,14 @@
+const shell = require('shelljs');
+
+function exec(cmd, options) {
+  const defaultOptions = { silent: true };
+  let output = shell.exec(cmd, { ...defaultOptions, ...(options || {}) });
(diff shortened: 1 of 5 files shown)


util/generic-functions.util.js
```diff
@@ -0,0 +1,14 @@
+const shell = require('shelljs');
+
+function exec(cmd, options) {
+  const defaultOptions = { silent: true };
+  let output = shell.exec(cmd, { ...defaultOptions, ...(options || {}) });
+  if (options && options.toString !== false) {
+    output = output.toString();
+    output = options.trim ? output.trim() : output;
```
ankushdharkar, line 8: NIT: Can be merged into one line

PR #63, opened 2020-08-23
change in social media icon- image wasn't showing

Harshi7016:


swarajpure (commented):
The twitter icon is still showing:
image

Can you please explain how is this PR needed?

Harshi7016:
Hi Swaraj,
I was observing the issue in the new page I created for member profile details page.
I tried different ways to mitigate that bug however only this was the option 
I can share a screenshot to show the same

swarajpure:
> 
> 
> Hi Swaraj,
> I was observing the issue in the new page I created for member profile details page.
> I tried different ways to mitigate that bug however only this was the option
> I can share a screenshot to show the s

Cool, got it! Can you please include a description of what you're trying to fix from the next time? Would be easier for others to understand 😄

Harshi7016:
Yeah, this was actually a minor fix kind of thing not related to issue I am working on but will help in fixing the issue I faced.
This is not the feature related fix :)
But next time , I will write my thoughts on why I am doing the same

ankushdharkar:
@Harshi7016 Do we need this PR?

Harshi7016:
I actually thought this bit was accepted,
I did this for members page , as twitter icon was not showing up 
Now that Nikhil has taken it, I am not sure , this would help or not
Ankush, adding / helped the issue 
If no one has seen the issue and if this would not help , then it can be deleted
I am bit confused here

Harshi7016:
I actually thought it was merged , however Swaraj has actually given an emoji instead

PR #74, opened 2020-11-22
Set cache for 12 Hrs

sumitd94:


components/member-profile/index.js
```diff
@@ -118,7 +118,7 @@ const Profile = (props) => {
                 height="16"
                 aria-hidden="true">
                 <path
-                  fill-rule="evenodd"
+                  fillRule="evenodd"
```
ankushdharkar, line 121: How do these changes help?

components/constants/max-age.js
```diff
@@ -0,0 +1 @@
+export const MAX_AGE = 43200;
```
ankushdharkar, line 1: Maybe make this a bit more descriptive? Max age of _what_?

ankushdharkar (changes requested):

components/constants/max-age.js
sumitd94: Yes Sure, I will make it CACHE_MAX_AGE

components/member-profile/index.js
sumitd94: Actually I am not sure why this changes came in, strange

ankushdharkar: Always scan your PR once, before sending in for a review :)

sumitd94: Noted :)

ankushdharkar (changes requested):
Please remove `fillRule` code that is not a part of this PR

ankushdharkar (approved):

components/member-profile/index.js
whyDontI: The error is because this issue,
https://stackoverflow.com/questions/39088153/how-do-i-get-rid-of-react-error-warnings-telling-me-to-change-svg-attributes-to

whyDontI: `eslint --fix` automatically makes it `fillRule`

PR #76, opened 2020-12-02
update navbar

gauravchauhan2403:
Navbar is updated.


components/navbar/index.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import classNames from './navbar.module.scss';
+
+const Navbar = () => {
```
swarajpure, line 4: It can be directed exported from here
`export default const Navbar.....`

swarajpure (changes requested):

components/navbar/index.js
gauravchauhan2403: can you explain it a little bit please. @swarajpure

swarajpure: You are defining the function on line number 4, right?
Also, it's the **only** function in the file (that is the reason why you have used `default` while exporting it)

When it's the **only** function to be exported, we can directly export it while defining it on line number 4 like:
`export default const Navbar = () =>`

In the current case, `Navbar` is being referenced twice on line number 4 for defining it and on line number 20 for exporting, which is kind of unnecessary and doesn't make much sense.

I see that this repo has been following what you have done (probably you have referred existing code and made your component accordingly). Let us discuss this on Discord 👍🏻

components/layout/index.js
```diff
@@ -1,6 +1,7 @@
 import Head from 'next/head';
 import PropTypes from 'prop-types';
 import Footer from 'components/footer';
+import Navbar from 'components/navbar';
 const Layout = (props) => {
   const { children, title, description } = props;
```
sumitd94, line 6: Since it receives only 3 props, we can destructure it while declaring the function
```javascript
const Layout = ({ title, description, children }) => {
```

sumitd94 (changes requested):

components/layout/index.js
gauravchauhan2403: @sumitd94  ok will do it.

gauravchauhan2403: @sumitd94 Please review it and suggest the changes if needed.

components/navbar/index.js
swarajpure: No need to do it as discussed on discord.

components/member-profile/index.js
```diff
@@ -143,7 +143,7 @@ const Profile = (props) => {
                 height="16"
                 aria-hidden="true">
                 <path
-                  fill-rule="evenodd"
+                  fillRule="evenodd"
```
sumitd94, line 146: Please revert this back to `fill-rule`

sumitd94 (changes requested):

components/member-profile/index.js
gauravchauhan2403: Change Done. Please Review.

components/member-profile/index.js
```diff
@@ -118,7 +118,7 @@ const Profile = (props) => {
...
```
sumitd94, line 121: It still shows `fillRule`
Please Remove this file from this commit

sumitd94 (changes requested):

sumitd94 (approved):

gauravchauhan2403:
@sumitd94 See it Now if it is resolved, Actually last time I Changed The fillRule at line no 146.

sumitd94:
> @sumitd94 See it Now if it is resolved, Actually last time I Changed The fillRule at line no 146.

Done, Thank you

Issue #104, opened 2021-01-20
Remove hardcoded image name 


Cartmanishere:
In `components/helper-functions/urls.js` I can see that we are constructing image url and have hardcoded `img.png`.

This can easily take a param for image name. We already have the image name in the member details in the website-static's `data.json`. 

I am working on this change and will raise a PR for this.

PR #105, opened 2021-01-23
Fix harcoded image name

Cartmanishere:
Closes: #104 

Use the image name returned in the member's data response to fetch the image.

Earlier the image name was hardcoded to `img.png`, so members had to add an image compulsorily named `img.png`.

Now, they just need to ensure that the image they add is correctly described in the `img` key in the member's `data.json`.

For example:

A new member adds their information in the `website-static` project as follows -- 

```
.
├── members
│   ├── anon
│   │   ├── data.json
│   │   └── profile.jpg
```

Then they would need to add it as following in `data.json` -- 

```
{
  "id": "anon"
  "img": "./profile.jpg",
}
```

### DIFF
--- a/helper-functions/urls.js
+++ b/helper-functions/urls.js
@@ -5,7 +5,7 @@ const PRbaseURL = 'https://staging-api.realdevsquad.com/pullrequests/user';
  *
  * @param {string} rdsId
  */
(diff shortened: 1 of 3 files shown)


sumitd94 (approved):
Thank you for the quick fix :)

PR #114, opened 2021-02-04
Click event added to member card

sanyogita14:
Profile page opens on clicking anywhere on the member card


sumitd94 (approved):
Thanks for the Fix!

PR #120, opened 2021-02-06
Eslint side-drawer error fixed

sanyogita14:
The issue for ESlint error in side-drawer component is fixed.


sumitd94 (approved):
Thanks for the fix.

PR #123, opened 2021-02-10
Show Spinner while the Route changes

biswajitripathy:


sumitd94 (approved):
Thank you for the implementation.

PR #125, opened 2021-02-18
changes done in the baseURL.

aman7870:
changed the base baseURL. imageURL is still same

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,6 +1,5 @@
 {
   "compilerOptions": {
-    "experimentalDecorators": true,
     "target": "es6",
     "module": "commonjs",
     "baseUrl": "./",
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "experimentalDecorators": true,
     "target": "es6",
     "module": "commonjs",
     "baseUrl": "./",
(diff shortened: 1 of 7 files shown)


aman7870:
did the change in the URL and modified according to the data.

pages/index.js
```diff
@@ -19,33 +19,32 @@ const Index = ({ membersArr, errorMsg }) => {
...
   try {
-    const res = await fetch(URL);
+    const res = await fetch(getMembersURL);
     if (res.status !== 200) {
       throw new Error(
         'There was some issues fetching the members, Please try again after some time'
       );
     }
-    let data = await res.json();
+    let { members } = await res.json();
```
sumitd94, line 32: We are not re-assigning `{ members }` here, let's declare it as `const`

pages/index.js
```diff
@@ -19,33 +19,32 @@ const Index = ({ membersArr, errorMsg }) => {
...
-        });
+    members.sort((a, b) => {
+      if (a.first_name < b.first_name) {
+        return -1;
       }
+      if (a.first_name > b.first_name) {
+        return 1;
+      }
+      return 0;
+    });
```
sumitd94, line 42: we can convert it into a single line 
```javascript
members.sort((a, b) => (a.first_name > b.first_name) ? 1 : -1);
```

Neat, clean & readable.

sumitd94 (changes requested):

jsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "experimentalDecorators": true,
```
sumitd94, line 3: Why is this changed?

sumitd94 (approved):
Thanks for the changes.

PR #156, opened 2021-04-13
Update github action to only use 14.x node version

Cartmanishere:


swarajpure (approved):
Thank you!

.github/workflows/continuous-integration.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x, 15.x]
+        node-version: [14.x]
```
swarajpure, line 19: Why was v15 added? @sumitd94

Issue #181, opened 2021-06-11
New Members section changes


sumitd94:
The following changes need to be done on the `new-members` section

- [x] Remove the hover effect from the member card
- [x] Reduce the size of each member card. It's too big right now
- [x] Every member card is not clickable, so remove the style `cursor: pointer`
- [x] There are too many members who do not have their `first_name` & `last_name` and on-screen it's showing it as `undefined undefined`, so only display those members who have their data set.
- [x] Change the design to adjust 6 members, or maybe 8 in one row (the idea is, new members should not be taking up this much space. Check and see if 8 looks good, or else will stick with 6)
- [x] Sort the new members by their firstName

PR #182, opened 2021-06-11
made necessary changes

Pavangbhat:
# Made necessary changes for new member section 
image

- [x] Remove the hover effect from the member card
- [x] Reduce the size of each member card. It's too big right now
- [x] Every member card is not clickable, so remove the style `cursor: pointer`
- [x] There are too many members who do not have their `first_name` & `last_name` and on-screen it's showing it as `undefined undefined`, so only display those members who have their data set.

#181


sumitd94 (approved):
Thanks for the quicky 🎉

PR #184, opened 2021-06-13
change in design for new member section

Pavangbhat:
- [x] Change the design to adjust 6 members, or maybe 8 in one row (the idea is, new members should not be taking up this much space. Check and see if 8 looks good, or else will stick with 6)
- [x] Sort the new members by their firstName
image

issue: #181


sumitd94 (approved):
Looks good. Thanks for the changes!

PR #190, opened 2021-06-23
added lang attribute to html element

abhinav1590:
i have added the necessary lang attribute to the html tag.


sumitd94 (approved):
Thanks for the changes!

PR #202, opened 2021-07-24
Added margins to the footer

rohandhar1234:
is that okay now?

### DIFF
--- a/components/footer/footer.module.scss
+++ b/components/footer/footer.module.scss
@@ -1,8 +1,7 @@
 .infoRepo {
-  display: flex;
   font-weight: 100;
   text-align: center;
-  padding:15px 0px 5px;
-  flex-shrink: 0;
+  display: flex;
   justify-content: center;
+  padding-top: 15px;
 }
--- a/components/footer/footer.module.scss
+++ b/components/footer/footer.module.scss
@@ -1,6 +1,6 @@
 .infoRepo {
   font-weight: 100;
-  margin: 15px auto;
+  margin: 20px auto;
   text-align: center;
   padding: 15px 0px;
   width: 100%;
--- a/components/footer/footer.module.scss
+++ b/components/footer/footer.module.scss
@@ -1,7 +1,10 @@
 .infoRepo {
   font-weight: 100;
   text-align: center;
-  display: flex;
-  justify-content: center;
-  padding-top: 15px;
+  position: absolute;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  height: 40px;
+  margin: 0 0 -40px 0;   
 }
--- a/components/footer/footer.module.scss
+++ b/components/footer/footer.module.scss
@@ -5,6 +5,6 @@
   left: 0;
   right: 0;
   bottom: 0;
-  height: 40px;
-  margin: 0 0 -40px 0;   
+  height: 45px;
+  margin: 0 0 -45px 0;   
 }
--- a/components/footer/footer.module.scss
+++ b/components/footer/footer.module.scss
@@ -1,10 +1,8 @@
 .infoRepo {
+  display: flex;
   font-weight: 100;
-  margin: 20px auto;
   text-align: center;
-  padding: 15px 0px;
-  width: 100%;
-  height: 2.5rem;
-  position: absolute;
-  bottom: 0;
+  padding:15px 0px 5px;
+  flex-shrink: 0;
+  justify-content: center;
 }
--- a/components/footer/footer.module.scss
+++ b/components/footer/footer.module.scss
@@ -1,6 +1,6 @@
 .infoRepo {
   font-weight: 100;
-  margin: 0;
+  margin: 10px auto 10px auto;
   text-align: center;
   padding: 15px 0px;
   width: 100%;
--- a/components/footer/footer.module.scss
+++ b/components/footer/footer.module.scss
@@ -1,6 +1,6 @@
 .infoRepo {
   font-weight: 100;
-  margin: 10px auto 10px auto;
+  margin: 15px auto;
   text-align: center;
   padding: 15px 0px;
   width: 100%;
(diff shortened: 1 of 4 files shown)


components/footer/footer.module.scss
```diff
@@ -1,6 +1,6 @@
 .infoRepo {
   font-weight: 100;
-  margin: 0;
+  margin: 10px auto 10px auto;
```
harshith-venkatesh, line 3: This could be reduced more,
top left bottom right
top,bottom left,right

harshith-venkatesh (changes requested):

components/footer/footer.module.scss
rohandhar1234: I could change left and right margins from auto to 0px and also adjust the top and bottom margins so that the footer issue is fixed. okay?

harshith-venkatesh: Did you go through the link?
What does it say?
Instead of repeating 10px twice, we can call it once if both are same values.

harshith-venkatesh: I am not sure about requirements, ask @sumitd94  please

rohandhar1234: yeah i studied the link and okay i will group in the same values together and pull again

harshith-venkatesh (approved):

rohandhar1234:
**BEFORE**
Screenshot (30)
**AFTER**
Screenshot (29)

pages/[id]/index.js
```diff
@@ -32,6 +32,7 @@ const MemberProfile = ({ imageLink, user, contributions, tasks, errorMessage })
         devUser={devUser}
         tasks={tasks}
       />
+        <br/><br/>
     </Layout>
```
Comment (harshith-venkatesh) [inline pages/[id]/index.js:36]: How is br / serving the purpose here

Comment (rohandhar1234) [reply pages/[id]/index.js:36]: The <br/> tag creates a line break after the content that has to be displayed before the footer.
Footer will always be displayed at the bottom because that's how it is structured ( in footer.sccss).
Footer is a part of <Layout/> and before footer is displayed at the bottom of the screen, I added these <br/> tags which created a line break after the member contents (such as, active tasks, contribution) got over, which naturally created a margin for the footer

Comment (correaswebert) [reply pages/[id]/index.js:36]: Please avoid doing that 😅  While it is a quick hack, it _might_ cause styling issues later on (in case our design guidelines change).
Do checkout [css-tricks.com](https://css-tricks.com/)

rohandhar1234:
**Members Page**

Screenshot (31)

**Member Content Page**

Screenshot (32)

rohandhar1234:
I got rid of the br/ tag and referred to the link @correaswebert shared. I used the sticky footer with the help of flexbox to handle this issue and then I aligned the text to center so that the footer is displayed in center from all devices.

components/footer/footer.module.scss
```diff
@@ -1,10 +1,8 @@
...
   font-weight: 100;
-  margin: 0;
   text-align: center;
-  padding: 15px 0px;
-  width: 100%;
-  height: 2.5rem;
-  position: absolute;
-  bottom: 0;
+  padding:15px 0px 5px;
+  flex-shrink: 0;
```
harshith-venkatesh, line 6: What is the advantage of flex-shrink here?

harshith-venkatesh (changes requested):

components/footer/footer.module.scss
```diff
@@ -1,10 +1,8 @@
...
+  display: flex;
   font-weight: 100;
-  margin: 0;
   text-align: center;
-  padding: 15px 0px;
-  width: 100%;
-  height: 2.5rem;
-  position: absolute;
-  bottom: 0;
+  padding:15px 0px 5px;
```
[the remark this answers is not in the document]
rohandhar1234: Yes, it was to align the footer according to the requirements

components/footer/footer.module.scss
rohandhar1234: There is not any utility of this code here, I added it because it was a part of the code snippet which I followed. I will make changes

rohandhar1234:
I have created a **scss** file for the layout component where the using the **flex grow** property, footer can be placed on the bottom of the page with and without the content there. I also adjusted the padding so that the footer always looks intact and in-place with proper formating.

rohandhar1234:
**Member Content Page**

Screenshot 2021-08-05 185203

**Members Page**

Screenshot 2021-08-05 185306






**Footer without any content**

withoutcontent

correaswebert (approved):
Fantastic 🥳

rohandhar1234:
**members page**

members-page

**member content page**

member-content

**without content**

without-content

**mobile-view**

mobile-view

components/layout/index.js
```diff
@@ -13,15 +13,17 @@ const Layout = ({ children, title, description }) => {
...
-      <Head>
-        <title>{title}</title>
-        <meta name="description" content={description} />
-        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-      </Head>
-      <Navbar drawerToggleClicked={sideDrawerToggleHandler} />
-      <SideDrawer open={showSideDrawer} close={sideDrawerToggleHandler} />
-      <div className="container">
-        <div>{children}</div>
+      <div>
```
harshith-venkatesh, line 16: I don't see any changes in this file, is this indentation changes

harshith-venkatesh:
This looks good, thanks for the screenshots 👍🏻

components/layout/index.js
rohandhar1234: I had created a new scss file in an earlier commit and so to revert back to normal changes without the creation of any component, I set the code to what it was initially. I removed these three lines of code and reset it to what it was at the very beginning.

ss

rohandhar1234: I think the PR you are reviewing is an old one. Please have a look at the latest PR for better understanding.

rohandhar1234: > I don't see any changes in this file, is this indentation changes

Looks like when I was reverting back to the original code of index.js (that is, to what it was originally), I forgot to remove the extra **div**  tags that I had created. I will apply the changes

harshith-venkatesh (approved):

lakshayman:
Looks fine to me

sumitd94 (approved):

PR #204, opened 2021-07-30
Basic setup for members context api

harshith-venkatesh:


constants/members-constant.js
```diff
@@ -0,0 +1,2 @@
+export const SET_MEMBERS = 'SET_MEMBERS';
+export const SET_ERRORS = 'SET_ERRORS';
```
sumitd94, line 2: The file can be renamed as `AppConstants`

store/members/members-context.js
```diff
@@ -0,0 +1,31 @@
...
+const initialState = {
+  membersArr: [],
+  newMembersArr: [],
+  errorMsg: ''
+};
+
+export const MembersProvider = ({ children }) => {
+  const [state, dispatch] = useReducer(membersReducer, initialState);
+  return (
+    <MembersContext.Provider value={{ membersState: state, membersDispatch: dispatch }}>
```
sumitd94, line 15: don't think so we explicitly have to specify `membersState & membersDispatch`.
Since it's in members' repo, it's well understood. What do you think?

sumitd94 (changes requested):
Few changes requested, please

store/members/members-context.js
harshith-venkatesh: Yeah, I was thinking of moving ahead with having single member context too as it would help accessing role of user or with getting logged in user details 
Something I was visioning in that fashion
If we are thinking in that direction having named states would be helpful right?
Open to suggestions

store/members/members-context.js
```diff
@@ -0,0 +1,27 @@
...
+  newMembersArr: [],
+  errorMsg: ''
+};
+
+export const MembersProvider = ({ children }) => {
+  const [state, dispatch] = useReducer(membersReducer, initialState);
+  return <MembersContext.Provider value={{ state, dispatch }}>{children}</MembersContext.Provider>;
+};
+
+export const useMembers = () => {
```
sumitd94, line 17: why are we naming this as `useMembers`? you know right in react if name something like `useSomeRandomTExt` it's considered as a custom hook and all the rules of hooks are need to be followed

store/members/members-context.js
```diff
@@ -0,0 +1,27 @@
...
+  const context = useContext(MembersContext);
```
harshith-venkatesh, line 18: I am using useContext here right, this is more of a custom hook created which can be accessed everywhere... That was my intention

sumitd94: it's not a custom hook, it's provided by react by default 🤔

harshith-venkatesh: useMembers is an custom hook created inside which I am accessing useContext which is provided by React.
Since we are using useContext here which is a react hook, I need to create custom hook useMembers right?

sumitd94 (approved):
Thanks for the changes, Great job!

PR #217, opened 2021-08-07
Fixed card click

lakshayman:
Added Cursor and Resolved Event Bubbling

### DIFF
--- a/components/member-profile/contribution/index.js
+++ b/components/member-profile/contribution/index.js
@@ -72,7 +72,7 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
         </div>
       </div>
       <div className={classNames.featureLink}>
-        <a href={featureUrl || prList[0]['url']} target="_blank" rel="noreferrer">
+        <a href={featureUrl || prList[0]['url']} onClick={(e)=>e.stopPropagation()} target="_blank" rel="noreferrer">
           Check out this feature in action
         </a>
       </div>
--- a/components/member-profile/contribution/index.js
+++ b/components/member-profile/contribution/index.js
@@ -41,10 +41,12 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
     }
   }
 
+  let url = featureUrl || prList[0]['url'];
+
   return (
     <div
-      className={classNames.contributionCard}
-      onClick={() => window.open(featureUrl || prList[0]['url'], '_blank')}
+      className={url && classNames.contributionCard}
+      onClick={() => url && window.open(url, '_blank')}
       onKeyDown={() => {}}
       aria-hidden="true">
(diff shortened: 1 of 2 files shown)


components/member-profile/contribution/index.js
```diff
@@ -42,7 +42,11 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
   }
 
   return (
-    <div>
+    <div
+      className={classNames.contributionCard}
+      onClick={() => window.open(featureUrl || prList[0]['url'], '_blank')}
```
ankushdharkar, line 47: I can see the logic for `featureUrl || prList[0]['url']` being repeated in two places
Can we move them to a common variable/function?

Also, what happens when we have no featureUrl and no prList elements? Do we have a runtime error?

ankushdharkar (changes requested):
Please check [this comment](https://github.com/Real-Dev-Squad/website-members/pull/217/files#r684667960)

components/member-profile/contribution/index.js
```diff
@@ -41,10 +41,12 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
     }
   }
 
+  let url = featureUrl || prList[0]['url'];
```
ankushdharkar, line 44: If there is no featureUrl, and no elements in prList, what happens?

ankushdharkar (changes requested):
Small doubt in comments

components/member-profile/contribution/index.js
```diff
@@ -41,8 +41,14 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
     }
   }
 
+  let url = featureUrl || prList[0]?.url;
```
sumitd94, line 44: why not use `const` for `url` rather than `let`? We are not re-assigning the value.

components/member-profile/contribution/index.js
```diff
@@ -41,8 +41,14 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
...
+
   return (
-    <div>
+    <div
+      className={url && classNames.contributionCard}
+      onClick={() => url && window.open(url, '_blank')}
+      onKeyDown={() => {}}
```
sumitd94, line 50: why is it blank? Don't you think we should be using `onKeyPress` and do the same thing as `onClick`?

components/member-profile/contribution/index.js
lakshayman: yes that's right changing it

components/member-profile/contribution/index.js
lakshayman: its the eslint requirement to put the onKeyDown
so i made it empty
yes I should put the same thing changing it

PR #219, opened 2021-08-08
Fix Card click

lakshayman:


harshith-venkatesh (approved):

components/member-profile/contribution/index.js
```diff
@@ -41,8 +41,14 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
...
   }
 
+  const url = featureUrl || prList[0]?.url;
+
   return (
-    <div>
+    <div
+      className={url && classNames.contributionCard}
+      onClick={() => url && window.open(url, '_blank')}
+      onKeyDown={() => url && window.open(url, '_blank')}
```
ankushdharkar, line 50:
```js 
onClick={goToURL}
onKeyDown={goToURL}
```

components/member-profile/contribution/index.js
```diff
@@ -68,9 +74,11 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
         </div>
       </div>
       <div className={classNames.featureLink}>
-        <a href={featureUrl || prList[0]['url']} target="_blank" rel="noreferrer">
-          Check out this feature in action
-        </a>
+        {url && (
+          <a href={url} onClick={(e) => e.stopPropagation()} target="_blank" rel="noreferrer">
```
ankushdharkar, line 78: We need to rethink this:

If the featureURL/URL is in members, we want to use the `<Link>`, otherwise we want to use `<a>`

ankushdharkar (changes requested):

components/member-profile/contribution/index.js
ankushdharkar: You should read up on client-side routing vs routing through `a` tags. This is a good starting point: https://nextjs.org/docs/routing/introduction

PR #223, opened 2021-08-09
Add wallet details in member profile

Vividh25:
The wallet details of the user are now visible on the member profile page, screenshots attached. 

Capture

Capturemob

Currently the wallet details are only visible when `dev=true` as per the requirements.

### DIFF
--- a/components/member-card/card.module.scss
+++ b/components/member-card/card.module.scss
@@ -79,6 +79,65 @@
   font-size: 0.88rem;
 }
 
+.showWallet, .showMemberWallet{
+  display: flex;
+  flex-wrap: wrap;
+  margin: 10px 0;
+  justify-content: center;
+}
+
+.walletContainer {
+  display: flex;
+  justify-content: space-between;
+  flex-wrap: wrap;
+  align-items: center;
+  border-radius: 2px;
+  width: 150px;
+  height: 70px;
+  padding: 10px 5px 2px;
+  margin-bottom: 8px;
+  margin-left: 6px;
+  background-color: #EEEEEE;
+
+  .coinData {
+    display: flex;
+    flex-direction: column;
+  }
+
+  .coinData>p {
+    height: 2px;
+  }
+
+  .coinData>p:first-child {
+    font-size: 12px;
+    font-weight: 500;
+    transform: translateY(-25px);
+    margin-right: 8px;
+  }
+
+  .coinData>p:last-child {
+    font-size: 20px;
+    font-weight: 700;
+  }
+
+  .coinDinero{
+    background-color: green;
+    width: 40px;
+    height: 40px;
+    border-radius: 50%;
+    box-shadow: rgb(0, 0, 0) 3px 3px 3px;
+    border: 1px solid green;
+  }
+  .coinNeelam{
+    background-color: blue;
+    width: 40px;
+    height: 40px;
+    border-radius: 50%;
+    box-shadow: rgb(0, 0, 0) 3px 3px 3px;
+    border: 1px solid blue;
+  }
+}
+
 @media only screen and (min-width: 1024px) {
   .showMemberSkills{
     justify-content: flex-start;
--- a/components/member-card/show-wallet.js
+++ b/components/member-card/show-wallet.js
@@ -1,0 +1,76 @@
+import classNames from 'components/member-card/card.module.scss';
+import { CURRENCIES, walletURL } from 'helper-functions/urls';
+import PropTypes from 'prop-types';
+import { useEffect, useState } from 'react';
+
+const Coins = (props) => {
+  const { coin, balance } = props;
+  return (
+    <div className={classNames.walletContainer}>
+      <div className={classNames.coinData}>
+        <p>{coin}</p>
+        <p>{balance}</p>
+      </div>
+      <div
+        className={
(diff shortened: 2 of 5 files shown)


components/member-card/card.module.scss
```diff
@@ -79,6 +79,58 @@
...
+    display: flex;
+    flex-direction: column;
+  }
+
+  .coinData>p {
+    height: 2px;
+  }
+
+  .coinData>p:first-child {
+    font-size: 12px;
```
swarajpure, line 105: Let's use `rem` for font-size please?
What if someone increases their system font size to say 32px? Your text would still be locked at 12px which would appear way too small

components/member-card/card.module.scss
```diff
@@ -79,6 +79,58 @@
...
+    font-weight: 500;
+    transform: translateY(-25px);
+    margin-right: 8px;
+  }
+
+  .coinData>p:last-child {
+    font-size: 20px;
```
swarajpure, line 112: re: font-size

components/member-card/show-wallet.js
```diff
@@ -0,0 +1,76 @@
...
+  useEffect(() => {
+    const fetchData = async () => {
+      await fetch(walletURL, { credentials: 'include' })
+        .then((res) => {
+          if (res.status >= 400 && res.status < 600) {
+            throw new Error('Bad response from server');
+          }
+          return res.json();
+        })
+        .then((resJson) => {
```
swarajpure, line 36: Why mix of async-await and .then?
Let's stick to async await please

components/member-card/show-wallet.js
```diff
@@ -0,0 +1,76 @@
...
+  }, []);
+
+  return (
+    <div className={show ? classNames.showMemberSkills : classNames.showSkills}>
+      {!isError && isLoading ? (
+        <p>Loading...</p>
+      ) : (
+        currencies.map((coin) => <Coins coin={coin.name} balance={coin.balance} key={coin.name} />)
+      )}
+    </div>
```
swarajpure, line 63: What about the case of `isError` true? We won't see anything in that case 😛

swarajpure (changes requested):
Wait a minute
- `/wallet` isn't a public route, it's behind auth. 
- Plus, it only returns user's own wallet. You can't see mine, I can't see yours, even when we have valid cookies

Need some clarification here

components/member-card/card.module.scss
Vividh25: Sure I'll add that.

components/member-card/show-wallet.js
Vividh25: Okay I'll change it

components/member-card/show-wallet.js
Vividh25: Yes😅 will add that.

components/member-card/card.module.scss
```diff
@@ -79,6 +79,58 @@
...
+    border: 1px solid green;
+  }
+  .coinNeelam{
+    background-color: blue;
+    width: 40px;
+    height: 40px;
+    border-radius: 50%;
+    box-shadow: rgb(0, 0, 0) 3px 3px 3px;
+    border: 1px solid blue;
+  }
```
sumitd94, line 131: Almost both of them have the same CSS except for the background.
Why not change the className to just `coin` and directly use the background color (inline style)

components/member-card/show-wallet.js
```diff
@@ -0,0 +1,76 @@
+import classNames from 'components/member-card/card.module.scss';
+import { CURRENCIES, walletURL } from 'helper-functions/urls';
+import PropTypes from 'prop-types';
+import { useEffect, useState } from 'react';
```
sumitd94, line 4: please move this to the top of the file.

components/member-card/show-wallet.js
```diff
@@ -0,0 +1,76 @@
...
+
+const Coins = (props) => {
+  const { coin, balance } = props;
+  return (
```
sumitd94, line 8: destructure the props directly on line no 6, and then we can omit the `return` statement and directly return it.

```javascript
const Coin = ({ coin, balance }) => (
 <div>...code</div>
);
```

components/member-card/show-wallet.js
```diff
@@ -0,0 +1,76 @@
...
+
+    fetchData();
+  }, []);
+
+  return (
+    <div className={show ? classNames.showMemberSkills : classNames.showSkills}>
+      {!isError && isLoading ? (
+        <p>Loading...</p>
+      ) : (
+        currencies.map((coin) => <Coins coin={coin.name} balance={coin.balance} key={coin.name} />)
```
sumitd94, line 61: Are `Dinero's, Neelam` coins or currencies?

components/member-card/card.module.scss
```diff
@@ -79,6 +79,58 @@
...
+
+  .coinData>p:first-child {
+    font-size: 12px;
+    font-weight: 500;
+    transform: translateY(-25px);
+    margin-right: 8px;
+  }
+
+  .coinData>p:last-child {
```
sumitd94, line 111: let's not apply CSS in this way, not a fan of it. Cant we give the last child any specific class name and then apply the css?

PR #226, opened 2021-08-16
Fix card click

lakshayman:
Fixed if  featureURL/URL is in members <Link> is used else <a>

### DIFF
--- a/components/member-profile/contribution/contribution.module.scss
+++ b/components/member-profile/contribution/contribution.module.scss
@@ -3,6 +3,9 @@
   display: flex;
   justify-content: space-between;
 }
+.contributionCard{
+  cursor: pointer;
+}
 .contributionCard:hover{
   transform: scale(1.03);
 }
--- a/components/member-profile/contribution/contribution.module.scss
+++ b/components/member-profile/contribution/contribution.module.scss
@@ -3,6 +3,9 @@
   display: flex;
   justify-content: space-between;
 }
+.contributionCard:hover{
+  transform: scale(1.03);
+}
 .leftSection{
   display: flex;
   flex-direction: column;
--- a/components/member-profile/contribution/index.js
+++ b/components/member-profile/contribution/index.js
@@ -72,7 +72,7 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
         </div>
       </div>
       <div className={classNames.featureLink}>
-        <a href={featureUrl || prList[0]['url']} target="_blank" rel="noreferrer">
+        <a href={featureUrl || prList[0]['url']} onClick={(e)=>e.stopPropagation()} target="_blank" rel="noreferrer">
           Check out this feature in action
         </a>
       </div>
--- a/components/member-profile/contribution/index.js
+++ b/components/member-profile/contribution/index.js
@@ -41,10 +41,12 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
     }
   }
 
+  let url = featureUrl || prList[0]['url'];
+
   return (
     <div
-      className={classNames.contributionCard}
-      onClick={() => window.open(featureUrl || prList[0]['url'], '_blank')}
+      className={url && classNames.contributionCard}
+      onClick={() => url && window.open(url, '_blank')}
       onKeyDown={() => {}}
       aria-hidden="true">
       <div className={classNames.contributionContainer}>
@@ -72,13 +74,11 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
         </div>
       </div>
       <div className={classNames.featureLink}>
-        <a
-          href={featureUrl || prList[0]['url']}
-          onClick={(e) => e.stopPropagation()}
-          target="_blank"
-          rel="noreferrer">
-          Check out this feature in action
-        </a>
+        {url && (
+          <a href={url} onClick={(e) => e.stopPropagation()} target="_blank" rel="noreferrer">
+            Check out this feature in action
+          </a>
+        )}
       </div>
       <hr className={classNames.line}></hr>
     </div>
--- a/components/member-profile/contribution/index.js
+++ b/components/member-profile/contribution/index.js
@@ -42,7 +42,11 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
   }
 
   return (
-    <div>
+    <div
+      className={classNames.contributionCard}
+      onClick={() => window.open(featureUrl || prList[0]['url'], '_blank')}
+      onKeyDown={() => {}}
+      aria-hidden="true">
       <div className={classNames.contributionContainer}>
         <div className={classNames.leftSection}>
           <h3 className={classNames.featureTitle}>{featureTitle}</h3>
--- a/components/member-profile/contribution/index.js
+++ b/components/member-profile/contribution/index.js
@@ -41,13 +41,13 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
     }
   }
 
-  let url = featureUrl || prList[0]?.url;
+  const url = featureUrl || prList[0]?.url;
 
   return (
     <div
       className={url && classNames.contributionCard}
       onClick={() => url && window.open(url, '_blank')}
-      onKeyDown={() => {}}
+      onKeyDown={() => url && window.open(url, '_blank')}
       aria-hidden="true">
       <div className={classNames.contributionContainer}>
         <div className={classNames.leftSection}>
--- a/components/member-profile/contribution/index.js
+++ b/components/member-profile/contribution/index.js
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import classNames from 'components/member-profile/contribution/contribution.module.scss';
 import PRLink from 'components/member-profile/contribution/pr-link';
 import { timeWas } from 'helper-functions/time-was';
+import Link from 'next/link';
 
 const renderPRLinks = (prList) =>
   prList.map(({ url }, index) => {
@@ -42,12 +43,14 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
   }
 
   const url = featureUrl || prList[0]?.url;
+  const gotoUrl = () => url && window.open(url, '_blank');
+  const urlObj = URL.parse(url)
 
   return (
     <div
       className={url && classNames.contributionCard}
-      onClick={() => url && window.open(url, '_blank')}
-      onKeyDown={() => url && window.open(url, '_blank')}
+      onClick={gotoUrl}
+      onKeyDown={gotoUrl}
       aria-hidden="true">
       <div className={classNames.contributionContainer}>
         <div className={classNames.leftSection}>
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
     "react-dom": "17.0.1",
     "react-hook-form": "^6.15.5",
     "sass": "^1.30.0",
+    "url": "^0.11.0",
```
ankushdharkar, line 26: Q: What does this give us, that's not already there in the [URL API](https://developer.mozilla.org/en-US/docs/Web/API/URL)?

components/member-profile/contribution/index.js
```diff
@@ -68,9 +113,16 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
...
       </div>
       <div className={classNames.featureLink}>
-        <a href={featureUrl || prList[0]['url']} target="_blank" rel="noreferrer">
-          Check out this feature in action
-        </a>
+        {url &&
+          (urlObj.host === 'members.realdevsquad.com' ? (
+            <Link href={urlObj.path}>
+              <button onClick={(e) => e.stopPropagation()}>Check out this feature in action</button>
+            </Link>
```
ankushdharkar, line 120: If we are not using the button, why render the markup here?

components/member-profile/contribution/index.js
```diff
@@ -68,9 +113,16 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
         </div>
       </div>
       <div className={classNames.featureLink}>
-        <a href={featureUrl || prList[0]['url']} target="_blank" rel="noreferrer">
-          Check out this feature in action
-        </a>
+        {url &&
+          (urlObj.host === 'members.realdevsquad.com' ? (
```
ankushdharkar, line 117: Nit: Abstract this into a `renderFeatureURL()` to decrease logic in return code path

components/member-profile/contribution/index.js
```diff
@@ -10,7 +12,51 @@ const renderPRLinks = (prList) =>
...
+            contribution={contribution}
+            fullName={fullName}
+            imageLink={imageLink}
+            devUser={devUser}
+            url={url}
+            urlObj={urlObj}
+          />
+        </div>
+      )}
+    </div>
```
ankushdharkar, line 53: Nit: Can we decrease the repeated lines here somehow?

Duplicated code can lead to software that is hard to understand and difficult to change. The Don't Repeat Yourself (DRY) principle states:

Every piece of knowledge must have a single, unambiguous, authoritative representation within a system.

When you violate DRY, bugs and maintenance problems are sure to follow. Duplicated code has a tendency to both continue to replicate and also to diverge (leaving bugs as two similar implementations differ in subtle ways).

ankushdharkar (changes requested):

package.json
lakshayman: i read about this but i thought those two are same.
changing it

components/member-profile/contribution/index.js
lakshayman: @ankushdharkar made the changes checkout the last commit

components/member-profile/contribution/index.js
```diff
@@ -10,11 +11,70 @@ const renderPRLinks = (prList) =>
...
+
+const ContributionCard = ({ contribution, fullName, imageLink, devUser, url, urlObj }) => {
+  const {
+    task: { title, startedOn, endsOn, status, purpose },
     prList
   } = contribution;
   const isTitleAvailable = title ? true : false;
   const featureTitle = isTitleAvailable ? title : prList[0].title;
+  const renderFeatureUrl = (url, urlObj) => {
+    if (url) {
```
harshith-venkatesh, line 61: I got to learn a lot from this PR, I had a question here , instead of using nested if's and passing return;
Could we go for conditional rendering here?
Like {URL &&(what we need to render here)}

What would be better practise here @ankushdharkar

components/member-profile/contribution/contribution.module.scss
```diff
@@ -40,6 +46,16 @@
...
   font-weight: 500;
 }
+.featureLink button{
+  background: none!important;
+  border: none;
+  padding: 0!important;
+  cursor: pointer;
+  text-decoration: none;
+  color: #041187;
+  font-size: 16px;
```
sumitd94, line 56: rem please

components/member-profile/contribution/contribution.module.scss
```diff
@@ -40,6 +46,16 @@
   font-size: 16px;
   font-weight: 500;
 }
+.featureLink button{
+  background: none!important;
```
sumitd94, line 50: please avoid using `!important`

components/member-profile/contribution/contribution.module.scss
```diff
@@ -40,6 +46,16 @@
...
+  border: none;
+  padding: 0!important;
```
sumitd94, line 52: re:

components/member-profile/contribution/contribution.module.scss
```diff
@@ -103,6 +119,9 @@
   .featureLink a{
     font-size: 14px;
   }
+  .featureLink button{
+    font-size: 14px;
+  }
```
sumitd94, line 124: we already have CSS defined for this button on line no 49 🤔

components/member-profile/contribution/index.js
```diff
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import classNames from 'components/member-profile/contribution/contribution.module.scss';
 import PRLink from 'components/member-profile/contribution/pr-link';
 import { timeWas } from 'helper-functions/time-was';
+import Link from 'next/link';
```
sumitd94, line 5: please move this to line no 1

components/member-profile/contribution/index.js
```diff
@@ -10,11 +11,70 @@ const renderPRLinks = (prList) =>
...
 const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
   const {
-    task: { title, startedOn, endsOn, status, purpose, featureUrl },
+    task: { featureUrl },
+    prList
+  } = contribution;
+  const url = featureUrl || prList[0]['url'];
+  const gotoUrl = () => url && window.open(url, '_blank');
+  const urlObj = url && new URL(url);
+  const card = () => (
```
sumitd94, line 20:
```suggestion
  const contributionCard = () => (
```

components/member-profile/contribution/index.js
```diff
@@ -10,11 +11,70 @@ const renderPRLinks = (prList) =>
...
+      contribution={contribution}
+      fullName={fullName}
+      imageLink={imageLink}
+      devUser={devUser}
+      url={url}
+      urlObj={urlObj}
+    />
+  );
+  const renderFeatureCard = () => {
+    if (urlObj?.host === 'members.realdevsquad.com') {
```
sumitd94, line 31: let's move `members.realdevsquad.com` to constants and then use that constat here.

components/member-profile/contribution/index.js
```diff
@@ -10,11 +11,70 @@ const renderPRLinks = (prList) =>
...
+      return (
+        <Link href={urlObj.pathname}>
+          <div className={url && classNames.contributionCard}>{card()}</div>
+        </Link>
+      );
+    } else {
```
sumitd94, line 37: we don't need an else here since we are returning in the `if` block.
so you can directly return the `div`

```javascript

if (condition) {
 return ()
}

return (the code that you are returning in the else block)
```

components/member-profile/contribution/index.js
```diff
@@ -10,11 +11,70 @@ const renderPRLinks = (prList) =>
...
+};
+
+const ContributionCard = ({ contribution, fullName, imageLink, devUser, url, urlObj }) => {
+  const {
+    task: { title, startedOn, endsOn, status, purpose },
     prList
   } = contribution;
   const isTitleAvailable = title ? true : false;
   const featureTitle = isTitleAvailable ? title : prList[0].title;
+  const renderFeatureUrl = (url, urlObj) => {
```
sumitd94, line 60: can you please add an empty line before this, becomes a bit readable

components/member-profile/contribution/index.js
```diff
@@ -87,4 +142,16 @@ Contribution.propTypes = {
...
+ContributionCard.propTypes = {
+  contribution: PropTypes.shape({
+    prList: PropTypes.array,
+    task: PropTypes.object
+  }).isRequired,
+  fullName: PropTypes.string.isRequired,
+  imageLink: PropTypes.string.isRequired,
+  devUser: PropTypes.bool,
+  url: PropTypes.string,
+  urlObj: PropTypes.object
```
sumitd94, line 154: default props for this?

components/member-profile/contribution/index.js
sumitd94: DO we really need to check this condition? I think the `url` will always be true? 🤔

components/member-profile/contribution/index.js
```diff
@@ -10,11 +11,70 @@ const renderPRLinks = (prList) =>
...
+            Check out this feature in action
+          </Link>
+        );
+      } else {
+        return (
+          <a href={url} onClick={(e) => e.stopPropagation()} target="_blank" rel="noreferrer">
+            Check out this feature in action
+          </a>
+        );
+      }
```
sumitd94, line 74: Few pointers here on the `if-else`

1. Move `members.realdevsquad.com` to a constat file and then use it here
2. Check [this](https://github.com/Real-Dev-Squad/website-members/pull/226/files#r695047922) comment

components/member-profile/contribution/index.js
```diff
@@ -10,11 +11,70 @@ const renderPRLinks = (prList) =>
...
+    }
+    return;
```
sumitd94, line 76: no need of this!

components/member-profile/contribution/index.js
lakshayman: @sumitd94 ankush gave me a case if there is no URL then what

components/member-profile/contribution/index.js
```diff
@@ -67,11 +126,7 @@ const Contribution = ({ contribution, fullName, imageLink, devUser }) => {
           )}
         </div>
       </div>
-      <div className={classNames.featureLink}>
-        <a href={featureUrl || prList[0]['url']} target="_blank" rel="noreferrer">
-          Check out this feature in action
-        </a>
-      </div>
+      <div className={classNames.featureLink}>{renderFeatureUrl(url, urlObj)}</div>
```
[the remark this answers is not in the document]
lakshayman: got it
changed

components/member-profile/contribution/index.js
```diff
@@ -8,13 +9,70 @@ const renderPRLinks = (prList) =>
     return <PRLink link={url} key={index} />;
   });
 
+const memberSiteHostName = 'members.realdevsquad.com';
```
sumitd94, line 12: constants should be capitalized

`HOST_NAME`
`membersSiteHostName` makes less sense, because we are already accessing members, so we need not explicitly say `membersSite`

Can you create a new file `appConstants` at the root and put the constants there and import it here?

sumitd94 (approved):

Issue #230, opened 2021-08-27
progressIndicator function doesn't have it's comments and it's doesn't return for one condition


takshch:
# Description 
progressIndicator function of helper-function category doesn't have it's comments. Also, it's doesn't return any value for one if-else condition which is required to return `class`

PR #231, opened 2021-08-27
Adds comments for progressIndicator helper-function and fix it's one if-else condition

takshch:
# Fixes - #230


ankushdharkar:
Not required. Moving to TS

PR #236, opened 2021-08-30
Sort members on the basis of parameters mentioned (WIP)

RohitKuwar:
@sumitd94 I have implemented the feature which searches the members by their names. Please review it.

### DIFF
--- a/components/members-list/index.js
+++ b/components/members-list/index.js
@@ -4,18 +4,32 @@ import MemberListItem from 'components/member-list-item';
 import classNames from 'components/members-list/member-list.module.scss';
 import { membersContext } from 'store/members/members-context';
 
-const MembersList = () => {
+const MembersList = ({ searchTerm }) => {
   const {
     state: { membersArr }
   } = membersContext();
   if (membersArr) {
+    console.log(membersArr);
     return (
       <div className={classNames.container}>
-        {membersArr.map((ele) => (
-          <React.Fragment key={ele['id']}>
-            <MemberListItem memberDetails={ele} />
-          </React.Fragment>
-        ))}
+        {membersArr.length > 0 &&
+          membersArr
+            .filter((ele) => {
+              if (searchTerm === '') {
+                return ele;
+              } else if (ele.first_name.toLowerCase().includes(searchTerm.toLocaleLowerCase())) {
+                return ele;
+              } else if (ele.last_name.toLowerCase().includes(searchTerm.toLocaleLowerCase())) {
+                return ele;
+              } else {
+                return null;
+              }
+            })
+            .map((ele) => (
+              <React.Fragment key={ele['id']}>
+                <MemberListItem memberDetails={ele} />
+              </React.Fragment>
+            ))}
       </div>
     );
   }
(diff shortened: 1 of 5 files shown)


components/search-members/search-members.module.scss
```diff
@@ -0,0 +1,12 @@
...
+  display: flex;
+  align-items: center;
+  padding: 6px;
+}
+.searchBox {
+  float: left;
+  padding: 4px;
+  border: 1px solid rgba(0, 0, 0, 0.397);
+  border-radius: 5px;
+  outline: none;
```
harshith-venkatesh, line 11: What is the purpose of removing outline here...
If it can be reverted it would be beneficial
https://betterprogramming.pub/a11y-never-remove-the-outlines-ee4efc7a9968

components/pages/index.js
```diff
@@ -1,16 +1,20 @@
+import { useState } from 'react';
 import PropTypes from 'prop-types';
 import MemberList from 'components/members-list';
 import classNames from 'components/pages/home-page.module.scss';
 import NewMemberList from 'components/new-member-list';
+import SearchMembers from 'components/search-members/index';
 
 const HomePage = () => {
+  const [searchTerm, setSearchTerm] = useState('');
```
harshith-venkatesh, line 9: Could you try exploring having context here, I feel it would be a better way to approach this

components/new-member-list/index.js
```diff
@@ -4,26 +4,40 @@ import React from 'react';
...
       <div className={classNames.container}>
-        {newMembersArr.map((ele) => (
-          <React.Fragment key={ele['id']}>
-            {(ele.first_name || ele.username) && <NewMemberListItem newMemberDetails={ele} />}
-          </React.Fragment>
-        ))}
+        {newMembersArr.length > 0 &&
+          newMembersArr
+            .filter((ele) => {
+              if (searchTerm === '') {
```
harshith-venkatesh, line 17: The same code is repeated twice and most of the if condition are returning the ele itself , could you look into having one util file and look for a way to make the function more concise

components/members-list/index.js
```diff
@@ -4,18 +4,32 @@ import MemberListItem from 'components/member-list-item';
...
     return (
       <div className={classNames.container}>
-        {membersArr.map((ele) => (
-          <React.Fragment key={ele['id']}>
-            <MemberListItem memberDetails={ele} />
-          </React.Fragment>
-        ))}
+        {membersArr.length > 0 &&
+          membersArr
+            .filter((ele) => {
```
harshith-venkatesh, line 17: same

harshith-venkatesh: The same code is repeated twice and most of the if condition are returning the ele itself , could you look into having one util file and look for a way to make the function more concise

harshith-venkatesh (changes requested):

components/search-members/search-members.module.scss
RohitKuwar: @harshith-venkatesh Thanks for this blog. I was not aware of the importance of an outline. I will keep the outline.

components/pages/index.js
RohitKuwar: ok

components/search-members/search-members.module.scss
swarajpure: Good one @harshith-venkatesh!
You could also share it in web/css channel on our discord server so more people get to know about it :)

PR #237, opened 2021-09-01
Generic Modal

sumitd94:
Created a Generic modal which will be used across the application!


components/UI/modal/modal.module.scss
```diff
@@ -0,0 +1,28 @@
+  
+.Modal {
+  position: fixed;
+  z-index: 500;
```
harshith-venkatesh, line 4: NIT: any specific reason for 500 as z-index

components/UI/modal/index.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classes from 'components/UI/modal/modal.module.scss';
+import Backdrop from 'components/UI/backdrop';
+
+const Modal = ({ show, modalClosed, children }) => (
+  <>
+    <Backdrop show={show} clicked={modalClosed} />
+    <div
+      className={classes.Modal}
```
harshith-venkatesh, line 10: are we following a convention here as I observe camelCase for `crossIcon` and `Modal`  here

harshith-venkatesh (commented):
Well written clean code @sumitd94.

components/UI/modal/modal.module.scss
```diff
@@ -0,0 +1,28 @@
...
+  background-color: white;
```
Pavangbhat, line 5: will background-color not be white by default  :thinking:

styles/global-styles.scss
```diff
@@ -1,3 +1,4 @@
+@import "https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.1/css/all.min.css";
```
swarajpure, line 1: How about using library instead? https://fontawesome.com/v5.15/how-to-use/on-the-web/using-with/react

Only those icons will be added to the final bundle which we'll use, right? Reducing the final bundle size compared to having to add whole of the css file from CDN to our bundle and then use only 1/2 out of it?

Please drop in your thoughts :)

components/UI/modal/modal.module.scss
swarajpure: What if we add fractional opacity here along with white color if we're going with it? or blue the background maybe? For blurring, I found this https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/blur()
but isn't supported by IE

components/UI/modal/index.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classes from 'components/UI/modal/modal.module.scss';
+import Backdrop from 'components/UI/backdrop';
+
+const Modal = ({ show, modalClosed, children }) => (
```
swarajpure, line 6: What if we add a default value for `show`?

swarajpure, line 6: Doesn't `modalClosed` sound weird because of past tense? How about `closeModal`?

swarajpure (changes requested):
Great job!

Need discussions on some part of implementation

styles/global-styles.scss
sumitd94: I initially thought of using this library, but for one icon I thought let's use it this way.

I am thinking of let's implement our own icons and use across RDS. Design team can help us.

Coming to increase in bundle size, don't worry about that.. next js has the feature of tree shaking, so in thr production build all the unused classes will be removed.

components/UI/modal/index.jsx
sumitd94: Dont think so we need a default value here, whenever we will be showing the modal, we will be passing the value as `true` (this true is nothing but the state of the modal, whether to show the modal or not) and accordingly the backdrop and transition will take place.

Once we implement this modal in our code, things will be more clear. Right now it's just the modal code.
Prakash will be going ahead and implement it in a feature.

components/UI/modal/modal.module.scss
sumitd94: Ah! 500 is not needed, 1 shoul just do the job. Thanks

components/UI/modal/modal.module.scss
sumitd94: This is needed. Understand what it's trying to do here.
Whenever we open the Modal, we show a backdrop that has a background black with a slight opacity.
Now, the modal is an overlay on top of that, so if we do not have to apply a background of white, this is how the Modal will look

Screenshot 2021-09-02 at 3 19 28 PM

And if we have a background of white, this is how it will look

Screenshot 2021-09-02 at 3 20 11 PM

Pavangbhat: Got it thank you for explaining 😃

Pavangbhat (dismissed):

PR #249, opened 2021-09-07
Implemented the generic modal

lakshayman:
Implemented generic modal on get intro button form on members profile page for devUser.


src/components/UI/modal/index.jsx
```diff
@@ -1,22 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import styles from 'components/UI/modal/modal.module.scss';
-import Backdrop from 'components/UI/backdrop';
+import styles from '@components/UI/modal/modal.module.scss';
+import Backdrop from '@components/UI/backdrop';
 
-const Modal = ({ show, closeModal, children }) => (
+const Modal = ({ show, closeModal, modalChildren, style }) => (
```
sumitd94, line 6: why `modalChildren`? it's already a Modal component and we do not have to explicitly say the children as `modalChildren`. just `children` should work fine.

src/components/UI/modal/index.jsx
```diff
@@ -1,22 +1,23 @@
...
+        ...style,
         transform: show ? 'translateY(0)' : 'translateY(-100vh)',
         opacity: show ? '1' : '0',
       }}
     >
       <span className={styles.crossIcon}>
         <i className="fa fa-times" aria-hidden="true" onClick={closeModal} />
       </span>
-      {children}
+      {modalChildren}
```
sumitd94, line 20: please change it to `children`

src/components/UI/modal/index.jsx
lakshayman: done changed

src/components/UI/modal/index.jsx
lakshayman: done

PR #250, opened 2021-09-07
disable cache for active tasks

### LINKED ISSUE #215: Do not cache Active tasks details on members contribution
Currently, we are caching the entire member's contribution details. We should avoid caching the active tasks so that it actively shows the changes being made to it.

Ishikabaid:
closes #215

### DIFF
--- a/src/pages/[id]/index.js
+++ b/src/pages/[id]/index.js
@@ -12,6 +12,7 @@ import Profile from '@components/member-profile';
 import NotFound from '@components/not-found-page';
 import Layout from '@components/layout';
 import { CACHE_MAX_AGE } from '@constants/cache-max-age.js';
+import { useEffect, useState } from 'react';
 
 const MemberProfile = ({
   imageLink,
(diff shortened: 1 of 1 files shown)


src/pages/[id]/index.js
```diff
@@ -20,6 +21,21 @@ const MemberProfile = ({
   tasks,
   errorMessage,
 }) => {
+  const [activeTasksData, setActiveTasksData] = useState([tasks]);
+  const getId = () => {
+    const router = useRouter();
+    const { id } = router.query;
+    return { id };
```
Comment (sumitd94) [inline src/pages/[id]/index.js:28]: Why are we returning the Id as an object? 🤔

Comment (Ishikabaid) [reply src/pages/[id]/index.js:28]: Actually I was a bit confused and thought we should return an object here...so we shouldn't be returning an object right??

Comment (akshay1502) [reply src/pages/[id]/index.js:28]: Since here, you are using `getId()` at `getActiveTasksURL()` function which requires `id` as argument and not `{ id }`. So we should return only id. @sumitd94 correct me if I am wrong.

Comment (sumitd94) [reply src/pages/[id]/index.js:28]: We shouldn't be returning an object here, just returning the id should work. Did you test whether it works like this? 🤔 
Let's return it as normal id without object
@Ishikabaid

Comment (Ishikabaid) [reply src/pages/[id]/index.js]: It wasn't giving any error in both the cases...Anyways I've returned Id without object.

sumitd94 (approved):

PR #255, opened 2021-09-23
Integrated Cloudinary images into member page

prakashchoudhary07:


src/helper-functions/urls.js
```diff
@@ -1,13 +1,22 @@
 const baseURL = `https://api.realdevsquad.com`;
 const imgBaseURL = `https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main`;
 const getMembersURL = `${baseURL}/members`;
+const cloudinaryImageUrl = `https://res.cloudinary.com/realdevsquad/image/upload`;
```
sumitd94, line 4: Can we please move this also to constants?

prakashchoudhary07: This file contains all the url's, I don't get the reason for moving this to another constant file and using it here again

sumitd94: Ah! sorry, my bad.

sumitd94 (approved):

src/pages/index.js
```diff
@@ -43,9 +48,12 @@ export async function getServerSideProps(context) {
     const { members } = await res.json();
 
     for (const memberData of members) {
+      const img_url = memberData.picture
+        ? getCloudinaryImgURL(memberData.picture.publicId, `w_${MAX_WIDTH}`)
+        : getImgURL(memberData.username, 'img.png');
```
sumitd94, line 53: Can we add this behind the feature flag? meaning this feature should only work if `dev=true`

harshith-venkatesh (approved):

sumitd94 (approved):

PR #260, opened 2021-10-02
Promote new-members to members

prakashchoudhary07:


src/components/member-role-update/index.js
```diff
@@ -0,0 +1,72 @@
...
+        >
+          Promote to Member
+        </button>
+        <br />
+        <p>{updateStatus}</p>
+      </>
+    );
+  };
+
+  return ReactDOM.createPortal(
```
harshith-venkatesh, line 56: Nice use of portal here for the modal

src/components/pages/index.js
```diff
@@ -1,16 +1,26 @@
+import { useContext } from 'react';
 import MemberList from '@components/members-list';
 import classNames from '@components/pages/home-page.module.scss';
 import NewMemberList from '@components/new-member-list';
+import UserProfile from '@components/user-profile';
+import MemberRoleUpdate from '@components/member-role-update';
+import UserContext from '@store/user/user-context';
 
 const HomePage = () => {
+  const { showMemberRoleUpdateModal } = useContext(UserContext);
```
harshith-venkatesh, line 10: I had a small confusion here, Is it better to define a function which returns `useContext(UserContext)` or call it in where ever we use it.

lakshayman: ya that's a nice question

src/store/user/user-context.js
```diff
@@ -0,0 +1,32 @@
...
+  };
+
+  return (
+    <UserContext.Provider value={initialUserContext}>
+      {children}
+    </UserContext.Provider>
+  );
+};
+
+export default UserContext;
```
lakshayman, line 32: Looks good to me

PR #262, opened 2021-10-18
Added mock wallet on member profile

Vividh25:
1. Created mock wallet data and used that to display the wallet details on every member's profile.
2. Made the requested changes on [this](https://github.com/Real-Dev-Squad/website-members/pull/223) PR.

### DIFF
--- a/src/components/member-card/card.module.scss
+++ b/src/components/member-card/card.module.scss
@@ -91,8 +91,7 @@
   width: 150px;
   height: 70px;
   padding: 10px 5px 2px;
-  margin-bottom: 8px;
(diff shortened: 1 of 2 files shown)


src/components/member-card/card.module.scss
```diff
@@ -82,6 +82,48 @@
...
+  margin: 0px 0px 8px 6px;
+  background-color: #eeeeee;
+
+  .coinData {
+    display: flex;
+    flex-direction: column;
+  }
+
+  .coinData > p {
+    height: 2px;
```
akshay1502, line 103: 2px? Or 20px?

Issue #240, opened 2021-09-03
Ankush can archive the old member profile from new members sections.  


MehulKChaudhari:
On the member's site, old members of RDS who are not active anymore are shown in the new member's section. 
So we have to add a feature in which Ankush can archive them from the new members section.

PR #263, opened 2021-10-18
Archive users/members from the members page

MehulKChaudhari:
This PR solves issue #240 

This PR should be merged only after merging PR #260


package.json
```diff
@@ -56,5 +56,8 @@
   },
   "engines": {
     "node": "14.x"
+  },
+  "optionalDependencies": {
```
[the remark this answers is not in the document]
MehulKChaudhari: Removed the dependency, resolved the error another way.

src/pages/index.js
```diff
@@ -64,7 +64,10 @@ export async function getServerSideProps(context) {
     membersArr.sort((a, b) => (a.first_name > b.first_name ? 1 : -1));
 
     const newMembersArr = membersArray.filter(
-      (person) => !person.isMember && person.first_name !== undefined
+      (person) =>
+        !person.isMember &&
+        person.first_name !== undefined &&
+        !person.archived_member
```
ankushdharkar, line 70: Why are we not keeping the field names consistent with camelCasing?

src/helper-functions/urls.js
```diff
@@ -43,6 +43,9 @@ const getActiveTasksURL = (rdsId) => `${baseURL}/tasks/${rdsId}?status=active`;
 const getAddMemberRoleURL = (rdsId) =>
   `${baseURL}/members/moveToMembers/${rdsId}`;
 
+const getArchiveMemberUrl = (rdsId) =>
```
ankushdharkar, line 46: Why is it URL above, but Url here?

src/components/member-role-update/index.js
```diff
@@ -26,6 +26,16 @@ const MemberRoleUpdate = () => {
       setUpdateStatus('Some error occured, please contact admin');
     }
   };
+  const archiveMember = async (user) => {
+    setIsUpdating(true);
+    const { status } = await archiveMember(user);
+    setIsUpdating(false);
```
ankushdharkar, line 32: Should we imperatively set the spinner logic, or have API state derive this data?

ankushdharkar (changes requested):

src/helper-functions/urls.js
MehulKChaudhari: Yeah, it should be a `URL `to keep code consistent.

src/pages/index.js
MehulKChaudhari: I didn't paid much attention, Kill keep in mind from now.

src/components/member-role-update/index.js
MehulKChaudhari: Yes we can do it API state derived, and that makes more sense.

MehulKChaudhari: https://github.com/Real-Dev-Squad/website-members/issues/272

ankushdharkar (approved):

PR #264, opened 2021-10-18
added dark mode feature

RohitKuwar:
A draft PR for the dark mode feature.

### DIFF
--- a/src/components/Dark-Theme/Themes.js
+++ b/src/components/Dark-Theme/Themes.js
@@ -6,14 +6,18 @@ export const lightTheme = {
   linkColor: '#0000FF',
   cardBoxShadow: '0 0 15px -7px rgba(0, 0, 0, 0.65)',
   prLink: '#455a64',
+  modalBackground: 'white',
+  modalHeader: 'blue',
 };
 export const darkTheme = {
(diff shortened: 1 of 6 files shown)


harshith-venkatesh:
@RohitKuwar , What is the importance of `styled-components` here?
Can we look into the way of building theme context without using `styled-component` ?

RohitKuwar:
`Styled-components` is basically `CSS-in-JS`. It allows you to write actual CSS to style React components.
Implementing dark mode with pure CSS makes it a complex thing because we have to think about all the elements styles that change conditionally. And it is difficult to maintain the code also. 
`Styled-components` solves the problem of difficulty in maintenance. Many times when we need to update a style, we end up breaking another style in another piece of code that has nothing to do with the first style. This makes maintenance much more costly than it should be and we need to keep digging a lot of code so we don't run the risk of breaking something that shouldn't be broken.
`Properties like -webkit, -moz-, -ms and -o- do not need to be entered anymore, as styled-components do this work automatically.`
It makes components less bulky.

RohitKuwar:
If it is possible to build `theme context` without using `styled-component` then I would like to try it also. Let me look into this.

PR #280, opened 2021-11-18
Show designers on home page

sumitd94:
This PR Contains changes to show designers on the home page.
This is behind the feature flag, to view the designers pass `dev=true` in the query params.

image

### DIFF
--- a/src/components/designers/mock-data.json
+++ b/src/components/designers/mock-data.json
@@ -1,0 +1,33 @@
+{
+  "designers": [
+    {
+      "id": "designerSoujanya",
+      "last_name": "Bagi",
+      "yoe": "2",
+      "incompleteUserDetails": false,
+      "roles": {
+        "member": true
+      },
+      "first_name": "Soujanya",
+      "company": "Real Dev Squad",
+      "linkedin_id": "soujanyabagi",
+      "username": "soujanya",
+      "isMember": true,
+      "img_url": ""
+    },
+    {
+      "id": "designerArushi",
+      "last_name": "",
+      "yoe": "2",
+      "incompleteUserDetails": false,
+      "roles": {
+        "member": true
+      },
+      "first_name": "Arushi",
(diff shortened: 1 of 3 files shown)


src/components/pages/index.js
```diff
@@ -20,6 +21,12 @@ const HomePage = () => {
         src="/images/Real-Dev-Squad@1x.png"
         alt="real-dev squad"
       />
+      {dev && (
+        <>
```
harshith-venkatesh, line 25: Why didn't we go with React.Fragment here?

harshith-venkatesh (approved):

src/components/designers/mock-data.json
```diff
@@ -0,0 +1,33 @@
...
+        "member": true
+      },
+      "first_name": "Arushi",
+      "company": "Real Deb Squad",
+      "username": "arushi",
+      "isMember": true,
+      "img_url": ""
+    }
+  ]
+}
```
prakashchoudhary07, line 33: Why are we mocking the user data? Don't we have their data in members API?
Designer's profile will be displayed on members page only, They won't have their own profile page, for now right?
Shouldn't we show their profile page also?

sumitd94: We will be showing their profile page as well.
Right now we do not have their data set up in the DB Yet. So @ankushdharkar suggested going with the mock data for designers.
We will be later setting things up for them from the ground up.

src/components/pages/index.js
sumitd94: Either way, it's fine 😅 
I personally don't like using `React.fragment` when we have the shorthand available.

src/components/designers/mock-data.json
prakashchoudhary07: ok, then it look good :+1:

prakashchoudhary07 (approved):

harshith-venkatesh (approved):

PR #291, opened 2021-12-27
Fetch members images from cloudinary

sumitd94:
### What is the change?

Added the logic to fetch members' images from Cloudinary and not from the static repository because we were making too many requests to fetch images for each member.

### Is it a bug?

- No

### \*Dev Tested?

- [ ] Yes
- [ ] No

#### Platforms

- [ ] Web

#### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox

Screenshot 2021-12-27 at 9 26 28 PM

Screenshot 2021-12-27 at 9 28 07 PM

### DIFF
--- a/src/components/member-card/index.js
+++ b/src/components/member-card/index.js
@@ -19,10 +19,6 @@ const Card = ({ developerInfo }) => {
   ];
   const fullName = `${`${first_name} ${last_name}`}`;
 
-  const brokenImageHandler = (e) => {
-    e.target.src = '/images/Avatar.png';
-  };
-
   const renderName = (userFullName, userName) =>
     userFullName.length > 20 ? userName : userFullName;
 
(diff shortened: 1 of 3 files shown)


lakshayman (dismissed):
Looks good to me

src/components/member-card/index.js
```diff
@@ -31,8 +27,7 @@ const Card = ({ developerInfo }) => {
       {!isMember && dev && <SuperUserOptions username={username} />}
       <motion.img
         layoutId={username}
-        src={`${img_url}?${Math.random() * 100}`}
-        onError={brokenImageHandler}
+        src={img_url}
         className={
           isMember ? classNames.imgContainer : classNames.imgContainerNewMember
```
prakashchoudhary07, line 32: I think we should handle the case if the Cloudinary image is broken or fails to load, what do you think :thinking:

prakashchoudhary07 (changes requested):

src/components/member-card/index.js
sumitd94: yes, you are right. I have reverted this.

prakashchoudhary07 (approved):

prakashchoudhary07:
Will we also change the user profile page images also (members/[id])?

Issue #294, opened 2022-01-02
[ESLint Warnings] Remove all the warnings


sumitd94:
Right now whenever we raise a PR, there is a lot of warning we see by ESlint in the PR.
We need to fix all those warnings

example PR for reference - https://github.com/Real-Dev-Squad/website-members/pull/291/files

If you scroll down, you will see all the warnings.

<img width="1408" alt="Screenshot 2022-01-02 at 1 09 48 PM" src="https://user-images.githubusercontent.com/29303618/147869344-1392815c-4b73-43a8-a199-94f44ab0ac9e.png">

Debarshi95:
Hey @sumitd94 , I would like to work on this one. Can you assign??

PR #295, opened 2022-01-02
Fix eslint warnings

Debarshi95:
### What is the change?
**This PR fixes the issue as mentioned in the issue https://github.com/Real-Dev-Squad/website-members/issues/294**

- Disabled **no-alert** rule.
- Renamed/Replaced unnamed functions with named functions or es6 arrow functions
- Updated the dependencies of UseEffect dependency array as per eslint warning

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -43,6 +43,7 @@ module.exports = {
     ],
     'react/jsx-props-no-spreading': 'off',
     'import/no-unresolved': [0, { caseSensitive: false }],
+    'no-alert': 'off',
   },
   settings: {
     'import/resolver': {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -43,6 +43,7 @@ module.exports = {
     ],
     'react/jsx-props-no-spreading': 'off',
     'import/no-unresolved': [0, { caseSensitive: false }],
+    'no-console': ['error', { allow: ['error'] }],
     'no-alert': 'off',
   },
   settings: {
--- a/src/helper-functions/fetch.js
+++ b/src/helper-functions/fetch.js
@@ -30,7 +30,6 @@ const fetch = async (
     });
     return response;
   } catch (err) {
-    console.error('Something went wrong. Please contact admin', err);
     return err;
   }
 };
--- a/src/helper-functions/fetch.js
+++ b/src/helper-functions/fetch.js
@@ -30,6 +30,7 @@ const fetch = async (
     });
     return response;
   } catch (err) {
+    console.error('Something went wrong. Please contact admin', err);
     return err;
   }
 };
--- a/.prettierrc
+++ b/.prettierrc
@@ -4,5 +4,6 @@
   "printWidth": 80,
   "singleQuote": true,
   "trailingComma": "es5",
-  "jsxBracketSameLine": false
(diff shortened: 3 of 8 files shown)


Debarshi95:
Hi @sumitd94, can you please kindly review if the changes are okay?

src/helper-functions/fetch.js
```diff
@@ -30,7 +30,6 @@ const fetch = async (
     });
     return response;
   } catch (err) {
-    console.error('Something went wrong. Please contact admin', err);
```
rohan09-raj, line 33: Why remove this line ?? I guess you can use alert here if this is causing es-lint errors or you can check for the rules to follow
https://docs.w3cub.com/eslint/rules/no-console

.eslintrc.js
```diff
@@ -43,6 +43,7 @@ module.exports = {
     ],
     'react/jsx-props-no-spreading': 'off',
     'import/no-unresolved': [0, { caseSensitive: false }],
+    'no-alert': 'off',
```
rohan09-raj, line 46: Instead of putting this, I think we can follow the rules of es-lint
https://docs.w3cub.com/eslint/rules/no-alert
What say ??

Debarshi95: Hey @rohan09-raj, couldn't find any exception to apply in the rule **no-alert**. Can you please guide how to solve the issue?

rohan09-raj: @Debarshi95 Sorry, I thought warnings can be removed by following the correct rules. I exactly do not know how it can be done, I think we can go with this rule.
Also, please add `'no-console': 'off' and put line 33 back in `fetch.js` which you have removed.
https://github.com/Real-Dev-Squad/website-members/pull/295/files/36ac38a3246fc8063f15702001ba7e9033487225#r778588984

src/helper-functions/fetch.js
rohan09-raj: Please add this back and set the rule `'no-console' : 'off'`

rohan09-raj (commented):
Please resolve the conflicts in `index.js` file

src/helper-functions/fetch.js
Debarshi95: @rohan09-raj , found an exception on the rule **no-console**: https://eslint.org/docs/rules/no-console. Should I apply this instead of disabling the entire rule?

rohan09-raj: Yes, I think we can use this.

.eslintrc.js
sumitd94: why is this required? can't we remove the alert's from the code?

rohan09-raj: @sumitd94 Yes, we can remove it, I forgot that Next JS already has features for popping up errors.

Debarshi95: @rohan09-raj May be we can show some toast notification or some kind of modal instead of showing alert?

rohan09-raj: I think @sumitd94 is the better person to suggest some good alternatives.

sumitd94: Let's leave it for now.
Will create toast service and track it in a different ticket

sumitd94 (approved):

PR #296, opened 2022-01-04
member-settings for admin activated on option hold

DashDeipayan:
### What is the change?

### Is it bug?

- Steps to repro
- Expected
- Actual

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [x] Web

#### Browsers

- [x] Chrome
- [x] Safari
- [ ] Firefox


package.json
```diff
@@ -21,17 +21,19 @@
     "axios": "^0.21.1",
     "cross-fetch": "^3.1.4",
     "framer-motion": "^3.2.1",
-    "next": "^10.0.4",
+    "next": "^12.0.7",
```
swarajpure, line 24: Woah, wait, why is your PR updating nextjs version?

package.json
```diff
@@ -21,17 +21,19 @@
...
+    "npm": "^8.3.0",
```
swarajpure, line 25: Looks like you've done `npm install npm` 😂

package.json
DashDeipayan: I have no idea. Don't even know if it's bad or good.

package.json
DashDeipayan: No idea...didn't do that for sure 🤣

package.json
ankushdharkar: ❌  Let's not be doing major upgrades like this. This is dangerous.

Good catch @swarajpure

Issue #274, opened 2021-11-03
Add fallback image for members/[id] page when user image is not available 


prakashchoudhary07:

rajat-mehra05:
@prakashchoudhary07  Is this issue still open ? Would like to work on this. Thank you.

PR #297, opened 2022-01-07
fix: members fallback image

rajat-mehra05:
This closes the Issue no. #274 

### What is the change?
 - In case the member profile pic is not found, the fallback image replaces the members pic.
 - Added `font-weight` in footer to make it readable.
 - Fix typo in `Readme.md`

### Before Change Screenshots

image

Footer:
image

###  After Change Screenshots

image

Footer:
image


README.md
```diff
@@ -78,7 +78,7 @@ export default BlogsPage;
 
 ```
 
-> Note: Do not create individual files in `/components` or `/pages`. Put every file in folders with appropiate names.
+> Note: Do not create individual files in `/components` or `/pages`. Put every file in folders with appropriate names.
```
akshay1502, line 81: Thanks for fixing the typo.

src/components/member-profile/index.js
```diff
@@ -358,7 +358,7 @@ const Profile = (props) => {
           <div className={classNames.memberDetails}>
             <motion.img
               layoutId={username}
-              src={imageLink.w_200}
+              src={imageLink.w_200 ?? '/icons/user.png'}
```
akshay1502, line 361: Not the ideal way to handle the error. You can check how we have handled the error in other files. [For reference](https://github.com/Real-Dev-Squad/website-members/blob/45dc8cae73e7b03152030e15230d0ecad0592bb6/src/components/member-card/index.js?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L35)

src/components/footer/footer.module.scss
```diff
@@ -1,5 +1,5 @@
 .infoRepo {
-  font-weight: 100;
+  font-weight: 400;
```
akshay1502, line 2: Could you add a before and after photo of it?

rajat-mehra05: Sure.

src/components/member-profile/index.js
rajat-mehra05: Sure.

src/components/member-profile/index.js
```diff
@@ -358,7 +362,8 @@ const Profile = (props) => {
           <div className={classNames.memberDetails}>
             <motion.img
               layoutId={username}
-              src={imageLink.w_200 ?? '/icons/user.png'}
+              src={imageLink.w_200 ?? '/images/Avatar.png'}
```
akshay1502, line 365: What if src has the wrong image URL(not null or undefined) which is not even existing? will it go for `images/Avatar.png`?

rajat-mehra05: Yes. If it's not able to fetch the member's image, it will go for Avatars.png. I have tested it for every scenario.

akshay1502: Testing every scenario is really great skill as a developer. But since we are already using the onError method, which will take care of all such scenarios.

akshay1502: Have you checked without the onError method?

rajat-mehra05: The `onError` method won't work if the user has given `src` as `imageLink.w_500`. It will produce the following output:

image


`onError` method will work in the following scenarios:
1. when `src === ""`
2. when `src === {imageLink.w_200}`

- But if we alter the numbers inside `imageLink`, the output shown above will be produced which we clearly don't want.
   May be this scenario was ignored. 
- `onError` method will come into effect for all other cases except this one. That's why we need to handle the `src` 
   attribute too.

akshay1502 (dismissed):

sumitd94 (dismissed):

harshith-venkatesh:
Do we need `package-lock.json` file to be merged from this PR as I am not observing any dependencies added here

rajat-mehra05:
> Do we need `package-lock.json` file to be merged from this PR as I am not observing any dependencies added here

No, we don't need to merge the `package-lock` file from this PR. I have removed it from my PR. You can review it. @harshith-venkatesh

akshay1502 (approved):

akshay1502 (approved):

akshay1502:
Good work @rajat-mehra05

rajat-mehra05:
> Good work @rajat-mehra05

Thanks @akshay1502

sumitd94 (approved):

Issue #311, opened 2022-01-15
[New Members] Do not fetch images from github static repo


sumitd94:
Right now we are trying to fetch images for all the members (including new members) from this static repo - https://github.com/Real-Dev-Squad/website-static

New members do not have their images in this repo and due to this, we are getting 404 errors in the console.

![Screenshot 2022-01-15 at 4 01 16 PM](https://user-images.githubusercontent.com/29303618/149618596-63d0180f-e91a-46e1-ac91-da63200e4648.png)

For new members we should not be fetching images at all and just display avatar.

abhay05:
Pls assign it to me

abhay05:
#314

PR #314, opened 2022-01-16
use placedholder image for new members

abhay05:
### What is the change?

Change to avoid fetching image from static site for new members
https://github.com/Real-Dev-Squad/website-members/issues/311#event-5899126198

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [x] Web

#### Browsers

- [x] Chrome
- [ ] Safari
- [x] Firefox

### DIFF
--- a/src/components/member-card/index.js
+++ b/src/components/member-card/index.js
@@ -31,7 +31,7 @@ const Card = ({ developerInfo }) => {
       {dev && <SuperUserOptions username={username} />}
       <motion.img
         layoutId={username}
-        src={`${img_url}?${Math.random() * 100}`}
+        src={isMember ? `${img_url}?${Math.random() * 100}` : '/images/Avatar.png'}
         onError={brokenImageHandler}
         className={
           isMember ? classNames.imgContainer : classNames.imgContainerNewMember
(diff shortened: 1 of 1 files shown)


src/components/member-card/index.js
```diff
@@ -31,7 +31,9 @@ const Card = ({ developerInfo }) => {
       {dev && <SuperUserOptions username={username} />}
       <motion.img
         layoutId={username}
-        src={`${img_url}?${Math.random() * 100}`}
+        src={
+          isMember ? `${img_url}?${Math.random() * 100}` : '/images/Avatar.png'
```
sumitd94, line 35: this is right but we have to avoid the fetching of images from here as well -  https://github.com/Real-Dev-Squad/website-members/blob/develop/src/pages/index.js#L58

abhay05: actually console errors are not appearing anymore, so why do have to refactor this snippet?

rds_before_change
as u can there is no 404 error , check the difference
rds_avoid_new_members_image_fetch_after_change

sumitd94 (approved):

Issue #330, opened 2022-02-17
Fetch cloudinary images only in members page with flag `dev=true`


pallabez:
- [ ] Make a new branch from main branch
- [ ] Set the `img_url` to store cloudinary image link when `dev=true` flag is on

pallabez:
@DeRaowl Assigned it to you

DeRaowl:
Hi @pallabez will raise PR by tomorrow

PR #334, opened 2022-02-20
Feat/img flag

DeRaowl:
Close
- https://github.com/Real-Dev-Squad/website-members/issues/330

### What is the change?

Checker flag added to render the image based on dev=true.

### Is it a bug? 
No

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [x] Web

#### Browsers

- [x] Chrome
- [ ] Safari
- [ ] Firefox

- Before
image

- After
image


src/pages/index.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
```
shubham-y, line 1: Can we try to avoid disabling this ESLint rule?

src/pages/index.js
```diff
@@ -50,12 +55,15 @@ export async function getServerSideProps(context) {
...
-        : getImgURL(username, 'img.png');
+
+      let img_url =
+        dev && picture
+          ? getCloudinaryImgURL(
+              picture.publicId,
+              `${WIDTH_200PX},${HEIGHT_200PX}`
+            )
+          : '';
+      if (!dev) img_url = getImgURL(username, 'img.png');
```
shubham-y, line 66: **Suggestion:** When dev is true, for users without Cloudinary images we should add a source of the default Avatar image instead of an empty string

shubham-y: Why is the logic of displaying member images different from the individual member page? 
Reference: [src/pages/[id]/index.js](https://github.com/Real-Dev-Squad/website-members/blob/5e26323e25110b1b13f6154c2c9f352d369c4edb/src/pages/%5Bid%5D/index.js#L76-L80)

shubham-y (changes requested):

src/pages/index.js
pallabez: For the second question as discussed: We are using the feature flag `dev=true` to let members know if their profile picture is properly set or not, so we don't want to fetch images from `website-static` repo and avoid confusion.

src/pages/index.js
DeRaowl: Removed the ESlint rule

src/pages/index.js
DeRaowl: @shubham-y default image added

shubham-y (dismissed):
LGTM👍

akshay1502 (dismissed):

src/pages/index.js
```diff
@@ -50,12 +54,15 @@ export async function getServerSideProps(context) {
     for (const memberData of members) {
       const { picture, username, isMember, first_name, archivedMember } =
         memberData;
-      const img_url = picture
-        ? getCloudinaryImgURL(
-            picture.publicId,
-            `${WIDTH_200PX},${HEIGHT_200PX}`
-          )
-        : getImgURL(username, 'img.png');
+
```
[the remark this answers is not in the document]
sumitd94: +1

sumitd94 (approved):

PR #353, opened 2022-04-19
fix-textChange

ShekharDhangar:
Changed Text from "New Members" to "New Users"

Visuals:
Before Changes:
image

After Changes:
image


DeRaowl:
@ShekharDhangar there are linting errors can you check?

PR #368, opened 2022-05-08
Implemented Configurable N-items list 

### LINKED ISSUE #162: Configurable N-items list
For the Noteworthy and All Tasks, provide a configurable variable that shows only N items, with a load more button on each list, which when clicked, loads and displays the next N items

palakgupta2712:
This feature closes #162.

This PR is to resolve conflicts of 
https://github.com/Real-Dev-Squad/website-members/pull/172

### DIFF
--- a/src/components/member-profile/contribution-type/contributions-type.module.scss
+++ b/src/components/member-profile/contribution-type/contributions-type.module.scss
@@ -42,3 +42,27 @@
     margin: 10px;
   }
 }
+
+.loadButton {
+  align-items: center;
+  background-color: white;
+  border-radius: 5px;
+  color: black;
+  cursor: pointer;
+  padding: 8px 8px;
+  font-size: 16px;
+  white-space: nowrap;
+  text-decoration: none;
+}
+
+.loadButton:hover {
+  color: white;
+  background-color: #041484;
+  transition: background-color 0.2s ease-in;
+  outline: #041484;
+}
+
+.loadDiv {
+  text-align: center;
+  margin-top: 1.5em;
+}
--- a/src/components/member-profile/contribution-type/index.js
+++ b/src/components/member-profile/contribution-type/index.js
@@ -1,9 +1,12 @@
+/* eslint-disable react/button-has-type */
 /* eslint-disable react/no-array-index-key */
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import classNames from '@components/member-profile/contribution-type/contributions-type.module.scss';
 import Contribution from '@components/member-profile/contribution/';
 import ActiveTask from '@components/member-profile/active-task';
+
+const INIT_CONTRIBUTIONS = 3;
 
 const renderContributions = (contributions, fullName, imageLink, devUser) =>
   contributions.map((noteWorthyContribution, index) => (
@@ -29,9 +32,17 @@ const ContributionType = (props) => {
   const { fullName, type, imageLink, contributions, devUser, tasks } = props;
 
   const [showMoreContent, setShowMoreContent] = useState(true);
+  const initCount =
+    devUser === true ? INIT_CONTRIBUTIONS : contributions.length;
+  const [count, setCount] = useState(initCount);
 
   const showMoreContentHandler = () => {
+    setCount(initCount);
     setShowMoreContent((prevstate) => !prevstate);
+  };
+
+  const loadMoreHandler = () => {
+    setCount((prevCount) => prevCount + INIT_CONTRIBUTIONS);
   };
 
   const showMoreContentClass = showMoreContent
@@ -54,7 +65,22 @@ const ContributionType = (props) => {
       <div className={showMoreContentClass}>
         {type !== 'Active tasks' ? (
           <div>
-            {renderContributions(contributions, fullName, imageLink, devUser)}
+            {renderContributions(
+              contributions.slice(0, count),
+              fullName,
+              imageLink,
+              devUser
+            )}
+            {count <= contributions.length && devUser && (
+              <div className={classNames.loadDiv}>
+                <button
+                  className={classNames.loadButton}
+                  onClick={loadMoreHandler}
+                >
+                  Load more
+                </button>
+              </div>
+            )}
           </div>
         ) : (
           <div>{renderActiveTasks(tasks)}</div>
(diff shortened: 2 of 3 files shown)


src/components/member-profile/contribution-type/index.js
```diff
@@ -29,11 +32,19 @@ const ContributionType = (props) => {
...
+  const initCount =
+    devUser === true ? INIT_CONTRIBUTIONS : contributions.length;
+  const [count, setCount] = useState(initCount);
 
   const showMoreContentHandler = () => {
+    setCount(initCount);
     setShowMoreContent((prevstate) => !prevstate);
   };
 
+  const loadMoreHandler = () => {
```
vvaibhavdesai, line 44: I think we can refactor this function to just a single line something like
`const loadMoreHandler=()=>setCount(prevCount=>pevCount+INIT_CONTRIBUTIONS)`

src/components/member-profile/contribution-type/index.js
```diff
@@ -54,7 +65,22 @@ const ContributionType = (props) => {
       <div className={showMoreContentClass}>
         {type !== 'Active tasks' ? (
           <div>
```
vvaibhavdesai, line 67: It would make it easier to read & manage later if we make this out as a component.

src/components/member-profile/contribution-type/contributions-type.module.scss
```diff
@@ -42,3 +42,27 @@
     margin: 10px;
   }
 }
+
```
vvaibhavdesai, line 45: why are we being particular with `px` and not using `rem`, for any specific reasons?

src/components/member-profile/contribution-type/contributions-type.module.scss
```diff
@@ -42,3 +42,27 @@
...
+  padding: 8px 8px;
+  font-size: 16px;
+  white-space: nowrap;
+  text-decoration: none;
+}
+
+.loadButton:hover {
+  color: white;
+  background-color: #041484;
+  transition: background-color 0.2s ease-in;
```
vvaibhavdesai, line 61: it would be great if we make sure we add ease-out transition to normal class as well of loadButton else if we unhover the style will look snappy

src/components/member-profile/contribution-type/index.js
```diff
@@ -54,7 +65,22 @@ const ContributionType = (props) => {
...
-            {renderContributions(contributions, fullName, imageLink, devUser)}
```
vvaibhavdesai, line 57: can we also take out this functions which are rendered directly with args something like
`const contributionInformation = {renderContributions(contributions, fullName, imageLink, devUser)}`
and then use the `contributionInformation` variable

src/components/member-profile/contribution-type/index.js
```diff
@@ -54,7 +65,22 @@ const ContributionType = (props) => {
...
+              contributions.slice(0, count),
+              fullName,
+              imageLink,
+              devUser
+            )}
+            {count <= contributions.length && devUser && (
+              <div className={classNames.loadDiv}>
+                <button
+                  className={classNames.loadButton}
+                  onClick={loadMoreHandler}
```
vvaibhavdesai, line 78: Appreciate that you haven't added inline functions, a good practice to follow as inline functions are anonymous. They are created every time the component renders. 😄

vvaibhavdesai (changes requested):
I think I have covered a major part of the review and have suggested some changes one thing which I noticed is wherever we have used map we are using index for keys. can we have something better to be used instead? Maybe it solves the purpose for now but I think we can fix this with some better solution.

src/components/member-profile/contribution-type/contributions-type.module.scss
palakgupta2712: we fix this in the next commit.

src/components/member-profile/contribution-type/index.js
```diff
@@ -54,7 +74,10 @@ const ContributionType = (props) => {
...
         {type !== 'Active tasks' ? (
           <div>
-            {renderContributions(contributions, fullName, imageLink, devUser)}
+            {contributionInformation}
+            {count <= contributions.length && devUser && (
+              <LoadMoreButton loadMoreHandler={loadMoreHandler} />
+            )}
           </div>
         ) : (
           <div>{renderActiveTasks(tasks)}</div>
```
vvaibhavdesai, line 83: instead of calling the function we ca store the value in a variable and use it

vvaibhavdesai (approved):

shreya-mishra (commented):
LGTM

shreya-mishra (approved):
LGTM

src/components/member-profile/load-more-btn/load-more-btn.module.scss
```diff
@@ -0,0 +1,24 @@
+.loadButton {
+  align-items: center;
+  background-color: white;
+  border-radius: 5px;
+  color: black;
+  cursor: pointer;
+  padding: 8px 8px;
+  font-size: 16px;
```
harshith-venkatesh, line 8: Wouldn't it be better to go with rem convention here?

src/components/member-profile/load-more-btn/index.js
```diff
@@ -0,0 +1,18 @@
...
+/* eslint-disable react/no-array-index-key */
+import React from 'react';
+import classNames from '@components/member-profile/load-more-btn/load-more-btn.module.scss';
+
+const LoadMoreButton = (props) => {
+  const { loadMoreHandler } = props;
+
+  return (
+    <div className={classNames.loadDiv}>
+      <button className={classNames.loadButton} onClick={loadMoreHandler}>
```
harshith-venkatesh, line 11: We could remove line number 1 by specifying button type here right?
https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/button-has-type.md

harshith-venkatesh: What is the reason behind no array index key here?
Could this be looked into please

src/components/member-profile/contribution-type/contributions-type.module.scss
```diff
@@ -39,6 +39,6 @@
 
 @media (max-width: 575.98px) {
   .container {
-    margin: 10px;
```
harshith-venkatesh, line 42: Wouldn't it be better to follow same convention with rest of files, i prefer rem as it has advantage with different resolution of screen than px, I would request you to do more research and come to collective conclusion

harshith-venkatesh (changes requested):

Issue #382, opened 2022-05-28
Multicolored new users


ankushdharkar:
Use userId hash to keep color same between reloads

![image](https://user-images.githubusercontent.com/1935403/170833481-277325b2-c8cf-4815-8da5-9a5b109ea71c.png)

vickydonor-99:
I would like to work on this issue.

vickydonor-99:
Now I am working on creating a hashing function for showing consistent user avataar's across refreshes

PR #396, opened 2022-07-21
Show colored UI avatars for new users

vickydonor-99:
### What is the change?
Changed ui-avatars of new members,  . [issue 382](https://github.com/Real-Dev-Squad/website-members/issues/382)

### Is it bug? No

- Steps to repro
- Expected
- Actual

### \*Dev Tested?

- [x] Yes
- [ ] No 

### \*Tested on:

browsers

#### Platforms

- [x] Web

#### Browsers

- [x] Chrome
- [x] Safari
- [x] Brave
- [ ] Firefox

## Before

Screenshot 2022-07-21 at 12 44 23 PM

##After

Screenshot 2022-07-21 at 12 44 40 PM

### DIFF
--- a/src/components/member-card/card.module.scss
+++ b/src/components/member-card/card.module.scss
@@ -3,9 +3,9 @@
   height: 12rem;
   border-radius: 50%;
   margin: 0px auto;
-  // FIXME: Remove Important
-  visibility: visible !important;
-  opacity: 1 !important;
+
+  visibility: visible;
+  opacity: 1;
 }
 .imgContainerNewMember {
   /* prettier-ignore */
@@ -70,6 +70,21 @@
 
 .settingsIcon {
   width: 1rem;
+}
+
+.new_member_img_absent_container {
+  height: 80px;
+  width: 80px;
+  border-radius: 56px;
+  margin: auto;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+.new_member_img_absent_container_text {
+  font-size: 36px;
+  letter-spacing: -1px;
+  font-weight: 900;
 }
 
 @media only screen and (min-width: 1024px) {
--- a/src/components/member-card/card.module.scss
+++ b/src/components/member-card/card.module.scss
@@ -72,17 +72,17 @@
   width: 1rem;
 }
 
-.new_member_img_absent_container {
-  height: 80px;
-  width: 80px;
-  border-radius: 56px;
+.newMemberImgAbsentContainer {
+  height: 5rem;
+  width: 5rem;
+  border-radius: 3.5rem;
   margin: auto;
   display: flex;
   align-items: center;
   justify-content: center;
 }
-.new_member_img_absent_container_text {
-  font-size: 36px;
+.newMemberImgAbsentContainerText {
+  font-size: 2.4rem;
   letter-spacing: -1px;
   font-weight: 900;
 }
--- a/src/components/member-card/index.js
+++ b/src/components/member-card/index.js
@@ -1,4 +1,4 @@
-import React, { useState, useRef, useEffect } from 'react';
+import React, { useState } from 'react';
 import { useRouter } from 'next/router';
 import classNames from '@components/member-card/card.module.scss';
 import { motion } from 'framer-motion';
@@ -12,8 +12,6 @@ const Card = ({ developerInfo, isOptionKey, colorCombination }) => {
   const { dev } = query;
   const { username, first_name, last_name, img_url, isMember } = developerInfo;
   const { color_primary, color_secondary } = colorCombination;
-  const cardRef = useRef();
-  const textRef = useRef();
 
   const socialMedia = [
     'twitter_id',
@@ -21,17 +19,6 @@ const Card = ({ developerInfo, isOptionKey, colorCombination }) => {
     'linkedin_id',
     'instagram_id',
   ];
-
-  useEffect(() => {
-    const cardElement = cardRef.current;
-    const textElement = textRef.current;
-    if (cardElement) {
-      cardElement.style.color = color_primary;
-      cardElement.style.border = `2px solid ${color_primary}`;
-      cardElement.style.backgroundColor = color_secondary;
-      textElement.innerText = `${first_name[0].toUpperCase()} ${last_name[0].toUpperCase()}`;
-    }
-  }, []);
 
   const fullName = `${`${first_name} ${last_name}`}`;
 
@@ -81,15 +68,18 @@ const Card = ({ developerInfo, isOptionKey, colorCombination }) => {
       )}
 
       {!isMember && (
-        <div
-          className={classNames.new_member_img_absent_container}
-          ref={cardRef}
+        <section
+          className={classNames.newMemberImgAbsentContainer}
+          style={{
+            color: color_primary,
+            border: `2px solid ${color_primary}`,
+            backgroundColor: color_secondary,
+          }}
         >
-          <p
-            ref={textRef}
-            className={classNames.new_member_img_absent_container_text}
-          />
-        </div>
+          <p className={classNames.newMemberImgAbsentContainerText}>
+            {`${first_name[0].toUpperCase()} ${last_name[0].toUpperCase()}`}
+          </p>
+        </section>
       )}
 
       <h2
--- a/src/components/member-card/index.js
+++ b/src/components/member-card/index.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useRef, useEffect } from 'react';
 import { useRouter } from 'next/router';
 import classNames from '@components/member-card/card.module.scss';
 import { motion } from 'framer-motion';
@@ -7,16 +7,32 @@ import PropTypes from 'prop-types';
 import ShowSkills from '@components/member-card/show-skills';
 import SuperUserOptions from '@components/member-card/super-user-options';
 
-const Card = ({ developerInfo, isOptionKey }) => {
+const Card = ({ developerInfo, isOptionKey, colorCombination }) => {
   const { query } = useRouter() || { query: { dev: false } };
   const { dev } = query;
   const { username, first_name, last_name, img_url, isMember } = developerInfo;
+  const { color_primary, color_secondary } = colorCombination;
+  const cardRef = useRef();
+  const textRef = useRef();
+
   const socialMedia = [
     'twitter_id',
     'github_id',
     'linkedin_id',
     'instagram_id',
   ];
+
+  useEffect(() => {
+    const cardElement = cardRef.current;
+    const textElement = textRef.current;
+    if (cardElement) {
+      cardElement.style.color = color_primary;
+      cardElement.style.border = `2px solid ${color_primary}`;
+      cardElement.style.backgroundColor = color_secondary;
+      textElement.innerText = `${first_name[0].toUpperCase()} ${last_name[0].toUpperCase()}`;
+    }
+  }, []);
+
   const fullName = `${`${first_name} ${last_name}`}`;
 
   const [showSettings, setShowSettings] = useState(false);
@@ -45,17 +61,37 @@ const Card = ({ developerInfo, isOptionKey }) => {
       {dev && (
         <SuperUserOptions username={username} showSettings={showSettings} />
       )}
-      <motion.img
-        layoutId={username}
-        src={
-          isMember ? `${img_url}?${Math.random() * 100}` : '/images/Avatar.png'
-        }
-        onError={brokenImageHandler}
-        className={
-          isMember ? classNames.imgContainer : classNames.imgContainerNewMember
-        }
-        alt={username}
-      />
+
+      {isMember && (
+        <motion.img
+          layoutId={username}
+          src={
+            isMember
+              ? `${img_url}?${Math.random() * 100}`
+              : '/images/Avatar.png'
+          }
+          onError={brokenImageHandler}
+          className={
+            isMember
+              ? classNames.imgContainer
+              : classNames.imgContainerNewMember
+          }
+          alt={username}
+        />
+      )}
+
+      {!isMember && (
+        <div
+          className={classNames.new_member_img_absent_container}
+          ref={cardRef}
+        >
+          <p
+            ref={textRef}
+            className={classNames.new_member_img_absent_container_text}
+          />
+        </div>
+      )}
+
       <h2
         className={
           isMember
@@ -92,10 +128,20 @@ Card.propTypes = {
     img_url: PropTypes.string.isRequired,
     isMember: PropTypes.bool.isRequired,
   }).isRequired,
+
   isOptionKey: PropTypes.bool,
+
+  colorCombination: PropTypes.shape({
+    color_primary: PropTypes.string.isRequired,
+    color_secondary: PropTypes.string.isRequired,
+  }),
 };
 Card.defaultProps = {
   isOptionKey: false,
+  colorCombination: {
+    color_primary: '#DB1212',
+    color_secondary: '#F88181',
+  },
 };
 
 export default Card;
--- a/src/components/new-members/index.js
+++ b/src/components/new-members/index.js
@@ -5,15 +5,8 @@ import styles from '@components/new-members/new-members.module.scss';
 import { membersContext } from '@store/members/members-context';
 import { searchMemberContext } from '@store/search-members/searchMembers-context';
 import { searchMembers } from '@helper-functions/search-members';
+import { cardColorArray } from '@constants/member-constants.js';
 
-const cardColorArray = [
-  { color_primary: '#DB1212', color_secondary: '#F88181' },
-  { color_primary: '#E9C46A', color_secondary: '#FFE094' },
-  { color_primary: '#F4A261', color_secondary: '#FFBF8C' },
-  { color_primary: '#2A9D8F', color_secondary: '#42E0CD' },
-  { color_primary: '#165692', color_secondary: '#3D8BD3' },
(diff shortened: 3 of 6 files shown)


src/components/member-card/index.js
```diff
@@ -45,17 +61,37 @@ const Card = ({ developerInfo, isOptionKey }) => {
...
+              : classNames.imgContainerNewMember
+          }
+          alt={username}
+        />
+      )}
+
+      {!isMember && (
+        <div
+          className={classNames.new_member_img_absent_container}
+          ref={cardRef}
```
ivinayakg, line 86: why create refs, cant we just simply pass on the data and style into the element.
any reasoning?

src/components/new-members/index.js
```diff
@@ -6,23 +6,42 @@ import { membersContext } from '@store/members/members-context';
 import { searchMemberContext } from '@store/search-members/searchMembers-context';
 import { searchMembers } from '@helper-functions/search-members';
 
+const cardColorArray = [
```
ivinayakg, line 9: push this code into a constants file or folder

src/components/new-members/index.js
```diff
@@ -6,23 +6,42 @@ import { membersContext } from '@store/members/members-context';
...
+  counter += 1;
+  if (counter > cardColorArray.length - 1) counter = 0;
+  return (
+    <div className={styles.containerForNewMember}>
+      {process.browser && (
+        <Card
+          developerInfo={newMember}
+          isMember={false}
+          isOptionKey={isOptionKey}
+          colorCombination={cardColorArray[prev]}
```
ivinayakg, line 31: instead of prev which is a non-reactive state, use a random function that generates the random no. between a range and pass that
```Math.floor(Math.random() * (cardColorArray.length))```

**Review by ivinayakg — COMMENTED** [review]
gave some suggestions, about the code which I belive can help make the code better

**Review by shreya-mishra — COMMENTED** [review]
LGTM :)

File: src/components/member-card/index.js (hunk 1 above)
Comment (vickydonor-99) [reply src/components/member-card/index.js:86]: Good question,  In react we avoid accessing the dom directly, with useRef we can access the virtual dom elements, thats why. I hope it clears the doubt

File: src/components/new-members/index.js (hunk 2 above)
Comment (vickydonor-99) [reply src/components/new-members/index.js:9]: sure

File: src/components/member-card/card.module.scss [hunk 4]
```diff
@@ -72,6 +72,21 @@
   width: 1rem;
 }
 
+.new_member_img_absent_container {
+  height: 80px;
```
DeRaowl, line 76: Can you make use of relative units?

src/components/member-card/card.module.scss
```diff
@@ -72,6 +72,21 @@
...
+  width: 80px;
+  border-radius: 56px;
+  margin: auto;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+.new_member_img_absent_container_text {
```
DeRaowl, line 84: className can be better

src/components/new-members/index.js
vickydonor-99: I would have done that In the first place, but since we decided on some kind of a pattern to be used I passed the colors in a sequence. But i dont get why `non-reactive state is important here`. We are not re-rendering the cards again once they are loaded.

src/components/member-card/index.js
```diff
@@ -7,16 +7,32 @@ import PropTypes from 'prop-types';
...
     'github_id',
     'linkedin_id',
     'instagram_id',
   ];
+
+  useEffect(() => {
+    const cardElement = cardRef.current;
+    const textElement = textRef.current;
+    if (cardElement) {
+      cardElement.style.color = color_primary;
```
DeRaowl, line 29: color_primary and color_seconday should also be as dependency to useEffect.

src/components/member-card/index.js
```diff
@@ -45,17 +61,37 @@ const Card = ({ developerInfo, isOptionKey }) => {
...
+            isMember
+              ? classNames.imgContainer
+              : classNames.imgContainerNewMember
+          }
+          alt={username}
+        />
+      )}
+
+      {!isMember && (
+        <div
```
DeRaowl, line 84: Can we make use of Section instead of div

src/components/member-card/index.js
```diff
@@ -45,17 +61,37 @@ const Card = ({ developerInfo, isOptionKey }) => {
...
+          className={classNames.new_member_img_absent_container}
```
DeRaowl, line 85: className can be better here

src/components/member-card/index.js
```diff
@@ -45,17 +61,37 @@ const Card = ({ developerInfo, isOptionKey }) => {
...
+          ref={cardRef}
+        >
+          <p
+            ref={textRef}
+            className={classNames.new_member_img_absent_container_text}
```
DeRaowl, line 90: classname can be better

src/components/new-members/index.js
```diff
@@ -6,23 +6,42 @@ import { membersContext } from '@store/members/members-context';
...
-      isOptionKey={isOptionKey}
-    />
-  </div>
-);
+const renderNewMember = (newMember, isOptionKey) => {
+  const prev = counter;
+  counter += 1;
+  if (counter > cardColorArray.length - 1) counter = 0;
+  return (
+    <div className={styles.containerForNewMember}>
```
DeRaowl, line 25: use Section instead of div

src/components/new-members/index.js
```diff
@@ -6,23 +6,42 @@ import { membersContext } from '@store/members/members-context';
...
+  );
+};
 
 const NewMemberList = ({ isOptionKey }) => {
   const {
     state: { newMembers },
   } = membersContext();
   const { searchTerm } = searchMemberContext();
   const filterMembers = searchMembers(newMembers, searchTerm);
+  // console.log(filterMembers);
```
DeRaowl, line 44: Remove console log

src/components/new-members/index.js
```diff
@@ -6,23 +6,42 @@ import { membersContext } from '@store/members/members-context';
...
 
+const cardColorArray = [
+  { color_primary: '#DB1212', color_secondary: '#F88181' },
+  { color_primary: '#E9C46A', color_secondary: '#FFE094' },
+  { color_primary: '#F4A261', color_secondary: '#FFBF8C' },
+  { color_primary: '#2A9D8F', color_secondary: '#42E0CD' },
+  { color_primary: '#165692', color_secondary: '#3D8BD3' },
+  { color_primary: '#264653', color_secondary: '#387892' },
+];
+let counter = 0;
```
DeRaowl, line 17: Why it is exposed to global scope?

src/components/member-card/index.js
```diff
@@ -7,16 +7,32 @@ import PropTypes from 'prop-types';
...
     'twitter_id',
     'github_id',
     'linkedin_id',
     'instagram_id',
   ];
+
+  useEffect(() => {
+    const cardElement = cardRef.current;
+    const textElement = textRef.current;
+    if (cardElement) {
```
DeRaowl, line 28: what if textElement is empty?

src/components/member-card/index.js
```diff
@@ -7,16 +7,32 @@ import PropTypes from 'prop-types';
...
+
   const socialMedia = [
     'twitter_id',
     'github_id',
     'linkedin_id',
     'instagram_id',
   ];
+
+  useEffect(() => {
+    const cardElement = cardRef.current;
```
DeRaowl, line 26: I don't think use of 2 useRef will be the better option. Can we make use of inline style instead of writing style in abstract way?

src/components/member-card/index.js
```diff
@@ -45,17 +61,37 @@ const Card = ({ developerInfo, isOptionKey }) => {
...
+          alt={username}
+        />
+      )}
+
+      {!isMember && (
+        <div
+          className={classNames.new_member_img_absent_container}
+          ref={cardRef}
+        >
+          <p
```
DeRaowl, line 88: instead of setting name in useEffect, can't we make us of helper function which extracts initials here?

DeRaowl (changes requested):

DeRaowl:
And if we look into screenshot colors dark, can we make us of subtle colors?

src/components/member-card/index.js
ivinayakg: That i know but the question is why you need to access the dom, can't we just pass down as JavaScript expression in the jsx straight

src/components/new-members/index.js
ivinayakg: It's a good practice in my opinion to be safe from the edge cases

vickydonor-99: can you elaborate more on this , help me get a better context

src/components/member-card/index.js
vickydonor-99: i can add them, but they aren't going to change after the initial render, any specific reason to add them?

src/components/member-card/card.module.scss
vickydonor-99: can u suggest a better classname?

ankushdharkar (changes requested):
Why has the package-lock updated?

The package-lock should not be changed, unless:
- The package.json is updated
- The PR is doing maintenance/upkeep of codebase

src/components/member-card/index.js
```diff
@@ -7,16 +7,32 @@ import PropTypes from 'prop-types';
 import ShowSkills from '@components/member-card/show-skills';
 import SuperUserOptions from '@components/member-card/super-user-options';
 
-const Card = ({ developerInfo, isOptionKey }) => {
+const Card = ({ developerInfo, isOptionKey, colorCombination }) => {
```
ankushdharkar, line 10: I am a little confused. We are updating the cards for the users. Why are we adding another prop in `member-card` for it?

src/components/member-card/index.js
```diff
@@ -7,16 +7,32 @@ import PropTypes from 'prop-types';
...
   const { query } = useRouter() || { query: { dev: false } };
   const { dev } = query;
   const { username, first_name, last_name, img_url, isMember } = developerInfo;
+  const { color_primary, color_secondary } = colorCombination;
+  const cardRef = useRef();
+  const textRef = useRef();
```
ankushdharkar, line 16: Why are we using refs here? Can we not derive this data that's used in the component?

src/components/member-card/index.js
```diff
@@ -7,16 +7,32 @@ import PropTypes from 'prop-types';
...
   ];
+
+  useEffect(() => {
+    const cardElement = cardRef.current;
+    const textElement = textRef.current;
+    if (cardElement) {
+      cardElement.style.color = color_primary;
+      cardElement.style.border = `2px solid ${color_primary}`;
+      cardElement.style.backgroundColor = color_secondary;
+      textElement.innerText = `${first_name[0].toUpperCase()} ${last_name[0].toUpperCase()}`;
```
ankushdharkar, line 32: Can the styling come from CSS modifiers instead?
We add the modifying class for the color(s), and the CSS included does the rest? Will make our code a lot more declarative in nature, and reduce the number of lines here.

src/components/member-card/index.js
```diff
@@ -7,16 +7,32 @@ import PropTypes from 'prop-types';
...
+    }
+  }, []);
```
ankushdharkar, line 34: React Hook useEffect has missing dependencies: 'color_primary', 'color_secondary', 'first_name', and 'last_name'.  Should they not be included?

src/components/new-members/index.js
ankushdharkar: I also want to remind @vickydonor-99, we want some sort of hashing function, that distributes the colors, but makes it such that the same color is always applied to the same username.

src/components/member-card/index.js
vickydonor-99: can u suggest a better classname?

src/components/member-card/index.js
vickydonor-99: thanks for the suggestion, made the change

src/components/member-card/card.module.scss
DeRaowl: newMembersImg or newMembersIcon should work

src/components/member-card/index.js
vickydonor-99: yeah looks like an overkill, thanks for the suggestion made the change

src/components/new-members/index.js
vickydonor-99: done

src/components/member-card/index.js
vickydonor-99: yeah looks like an overkill, thanks for the suggestion made the change

src/components/member-card/card.module.scss
vickydonor-99: sure

src/components/member-card/index.js
vickydonor-99: sure

src/components/member-card/index.js
vickydonor-99: update: removed useEffect altogether

src/components/new-members/index.js
vickydonor-99: okay

src/components/member-card/index.js
vickydonor-99: removed the refs

src/components/member-card/index.js
vickydonor-99: for now i have added inline styles, later might make the suggested change, once I get the idea about the hashing function

src/components/member-card/index.js
vickydonor-99: removed the useEffect, not required now. Also after the initial render, the components are not updating so i would not have mattered

src/components/member-card/card.module.scss
vickydonor-99: The current classname suggest that the image for the user is not present hence we are showing their initials, the suggested classname doesn't do that

src/components/member-card/index.js
vickydonor-99: logic for new-member and new-member was present in the same card to begin with

src/components/new-members/index.js
vickydonor-99: yes @ankushdharkar, haven't got a clear idea about it yet

vickydonor-99:
> Why has the package-lock updated?
> 
> The package-lock should not be changed, unless:
> 
> * The package.json is updated
> * The PR is doing maintenance/upkeep of codebase

ivinayakg:
> > Why has the package-lock updated?
> > The package-lock should not be changed, unless:
> > 
> > * The package.json is updated
> > * The PR is doing maintenance/upkeep of codebase
> 
> I have not updated the `package.json`, not sure why this happened

@vickydonor-99 
When we do npm i, it install the packages based on the package.json which doesn't hold the exact package version which is being used in project. 
And therefore it may change the package-lock.json to avoid this next time 
Do install based on package-lock.json
Or next don't add package.json file in got stage

vickydonor-99:
> > > Why has the package-lock updated?
> > > The package-lock should not be changed, unless:
> > > 
> > > * The package.json is updated
> > > * The PR is doing maintenance/upkeep of codebase
> > 
> > 
> > I have not updated the `package.json`, not sure why this happened
> 
> @vickydonor-99 When we do npm i, it install the packages based on the package.json which doesn't hold the exact package version which is being used in project. And therefore it may change the package-lock.json to avoid this next time Do install based on package-lock.json Or next don't add package.json file in got stage

src/components/member-card/index.js
```diff
@@ -45,17 +48,40 @@ const Card = ({ developerInfo, isOptionKey }) => {
...
+          src={
+            isMember
+              ? `${img_url}?${Math.random() * 100}`
+              : '/images/Avatar.png'
+          }
+          onError={brokenImageHandler}
+          className={
+            isMember
+              ? classNames.imgContainer
+              : classNames.imgContainerNewMember
```
sumitd94, line 64: do we really need this check? given the fact this section only renders when `isMember` is true 🤔

src/components/member-card/index.js
```diff
@@ -45,17 +48,40 @@ const Card = ({ developerInfo, isOptionKey }) => {
...
+            color: color_primary,
+            border: `2px solid ${color_primary}`,
+            backgroundColor: color_secondary,
+          }}
+        >
+          <p className={classNames.newMemberImgAbsentContainerText}>
+            {`${first_name[0].toUpperCase()} ${last_name[0].toUpperCase()}`}
+          </p>
+        </section>
+      )}
```
sumitd94, line 83: NIT

Why not

```javascript
{isMember ? <render something> : <render something>}
```

src/components/member-card/index.js
```diff
@@ -92,10 +118,20 @@ Card.propTypes = {
...
+  colorCombination: PropTypes.shape({
+    color_primary: PropTypes.string.isRequired,
+    color_secondary: PropTypes.string.isRequired,
+  }),
 };
 Card.defaultProps = {
   isOptionKey: false,
+  colorCombination: {
+    color_primary: '#DB1212',
+    color_secondary: '#F88181',
```
sumitd94, line 133: can we move these constants to a constat file and then import them here and use it? In the future, we might need these default colors elsewhere also. What do you think?

src/components/member-card/index.js
vickydonor-99: thanks for the suggestion, we don't need it

src/components/member-card/index.js
vickydonor-99: yup that works as well thanks

Issue #367, opened 2022-05-06
Disallow profile page of non members to be seen by non super user. 


pallabez:
- https://members.realdevsquad.com/dhanush_kumar027

![image](https://user-images.githubusercontent.com/63191499/167204190-be6c42b2-a724-4657-99c1-1d4e8f03a059.png)

PR #403, opened 2022-08-07
add Checks for Super User while viewing Non Members Pages

heyrandhir:
### What is the change?
Issue: https://github.com/Real-Dev-Squad/website-members/issues/367
- [x] Adds a check for Super User while viewing pages of Non-Members. Only Super User can view the Non-Members Page.
- [x] Adds a check for users that do not exist.

### Is it a bug?
 Yes

- Steps to repro
- Expected
- Actual

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [x] Web

#### Browsers

- [x] Chrome
- [ ] Safari
- [ ] Firefox

Before Anyone can view non-member profile page
image

After:
Non super-users cannot access non-member profile
image

### DIFF
--- a/src/pages/[id]/index.js
+++ b/src/pages/[id]/index.js
@@ -19,23 +19,37 @@ import {
   HEIGHT_40PX,
 } from '@constants/profile-image';
 import { useEffect, useState } from 'react';
+import { userContext } from '@store/user/user-context';
 
 const MemberProfile = ({ imageLink, user, contributions, errorMessage }) => {
+  const { isSuperUserMode } = userContext();
   const [activeTasksData, setActiveTasksData] = useState([]);
   const router = useRouter();
   const { id } = router.query;
-  useEffect(() => {
-    (async () => {
-      const tasksURL = getActiveTasksURL(id);
-      const tasksResponse = await fetch(tasksURL);
-      const { tasks } = await tasksResponse.data;
-      setActiveTasksData(tasks);
-    })();
-  }, [id]);
 
   if (errorMessage) {
     return <NotFound errorMsg={errorMessage} />;
   }
+  if (!user.roles?.member && !isSuperUserMode) {
+    return (
+      <NotFound errorMsg="You dont have the necassary permissions to view this page." />
+    );
+  }
+
+  const fillActiveTasksArray = async () => {
+    const tasksURL = getActiveTasksURL(id);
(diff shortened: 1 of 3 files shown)


prakashchoudhary07 (approved):
LGTM

Maheima:
LGTM

src/pages/[id]/index.js
```diff
@@ -19,23 +19,35 @@ import {
...
+  const fillActiveTasksArray = async () => {
+    const tasksURL = getActiveTasksURL(id);
+    try {
+      const tasksResponse = await fetch(tasksURL);
+      const { tasks } = await tasksResponse.data;
+      setActiveTasksData(tasks);
+    } catch {
+      setActiveTasksData([]);
+    }
+  };
```
Comment (harshith-venkatesh) [inline src/pages/[id]/index.js:46]: Could you check more on useCallback?
and we could go ahead and add `fillActiveTasksArray` in the dependency there

Comment (harshith-venkatesh) [reply src/pages/[id]/index.js:46]: https://dmitripavlutin.com/dont-overuse-react-usecallback/

I love these blogs along with official docs ofcourse :)

src/pages/[id]/index.js
```diff
@@ -19,23 +19,35 @@ import {
...
-      const { tasks } = await tasksResponse.data;
-      setActiveTasksData(tasks);
-    })();
-  }, [id]);
 
   if (errorMessage) {
     return <NotFound errorMsg={errorMessage} />;
   }
+  if (!user.roles?.member && !isSuperUserMode) {
+    return <NotFound errorMsg="Unauthorized to view this page." />;
```
Comment (harshith-venkatesh) [inline src/pages/[id]/index.js:34]: Is this better to save this as constants if we are planning to utilize it in different pages?

harshith-venkatesh (changes requested):

src/pages/[id]/index.js
Comment (heyrandhir) [reply src/pages/[id]/index.js]: Thanks for the blog Harshith. It was indeed helpful. I have added the fillActiveTasksArray in the dependency array & wrapped fillActiveTasksArray in useCallBack.

src/pages/[id]/index.js
Comment (heyrandhir) [reply src/pages/[id]/index.js]: yes, it's better to add this as a constant. I have created a new constant for the same now :).

PR #358, opened 2022-04-20
update: added volta

vvaibhavdesai:
- added volta
- updated package.json with node & yarn version
- removed package-lock### What is the change?

Provide a small description of what did you change and provide the reference to the issue ticket.

### Is it bug?

- Steps to repro
- Expected
- Actual

### \*Dev Tested?

- [ ] Yes
- [ ] No

#### Platforms

- [ ] Web

#### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox

### Before / After Change Screenshots

> For visual or interaction changes. Can be video / screenshot.

PR #405, opened 2022-08-11
Fix/volta addition

kshitij430:
**Fixed Merge conflicts occurring in volta/addition branch.**

### What is the change?

### Is it bug?

- Steps to repro
- Expected
- Actual

### \*Dev Tested?

- [ ] Yes
- [ ] No

#### Platforms

- [ ] Web

#### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox


package.json
```diff
@@ -63,5 +63,9 @@
   },
   "engines": {
     "node": "14.x"
+  },
+  "volta": {
+    "node": "14.19.1",
+    "yarn": "1.19.2"
```
ankushdharkar, line 69: Can we have this at the following values please?

```
    "node": "16.16.0",
    "yarn": "1.22.17"
```

ankushdharkar (changes requested):
Generally changes look good, but there are a few more changes we need:

1. Readme documentation is not updated to provide instructions for using yarn and volta, instead of npm
2. package-lock is no longer used, so it should be removed and added to .gitignore
3. Most of yarn commands don't need 'run'. Please update those.

Refer to these recent changes for the calendar site which did something similar: https://github.com/Real-Dev-Squad/website-calendar/pull/42/files

package.json
kshitij430: Hi, Ankush 
I have a question, the branch volta-addition which is resolving PR #358 has package.json file with different dependency versions and our develop branch is having package.json with different ones. this is causing yarn.lock file to change after installing necessary modules. 
for example ->  "next": "^10.0.4" (In volta-addition branch) and "next": "^12.1.0"(In develop branch),
Solution -> after executing yarn, the changes in my yarn.lock file can fix this, can I also commit those changes?

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules/
 .next/
-out/
\ No newline at end of file
+out/
+package-lock.json
```
sumitd94, line 4:
```suggestion
package-lock.json

```

ankushdharkar (approved):
Thank you! This is great.

ankushdharkar (approved):

sumitd94 (approved):

PR #410, opened 2022-08-15
Profile page of nonMembers only to be visible to SuperUsers.

heyrandhir:
### What is the change?
Issue: https://github.com/Real-Dev-Squad/website-members/issues/367

- Only Super Users should be able to view the Non-Members Profile page.

### Is it bug?
Yes

- Steps to repro
- Expected
- Actual

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [x] Web

#### Browsers

- [x] Chrome
- [ ] Safari
- [ ] Firefox

### DIFF
--- a/src/components/user-profile/index.js
+++ b/src/components/user-profile/index.js
@@ -26,7 +26,7 @@ const UserProfile = () => {
       }
     }
     getUserProfile();
-  }, [setUser, setIsSuperUser]);
+  }, [setUser, setIsSuperUser, setApiCalledToVerifySuperUser]);
 
   const showSuperUserOptions = () => {
     return (
--- a/src/components/user-profile/index.js
+++ b/src/components/user-profile/index.js
@@ -12,6 +12,7 @@ const UserProfile = () => {
     setIsSuperUser,
     isSuperUserMode,
     setIsSuperUserMode,
+    setApiCalledToVerifySuperUser,
   } = userContext();
 
   useEffect(() => {
@@ -20,7 +21,8 @@ const UserProfile = () => {
       setIsLoading(false);
       setUser(data);
       if (data) {
-        setIsSuperUser(data.roles.super_user);
+        setIsSuperUser(Boolean(data.roles?.super_user));
+        setApiCalledToVerifySuperUser(true);
       }
     }
     getUserProfile();
--- a/src/components/user-profile/index.js
+++ b/src/components/user-profile/index.js
@@ -9,15 +9,15 @@ const UserProfile = () => {
     isSuperUser,
     isSuperUserMode,
     setIsSuperUserMode,
-    verifySuperUser,
+    setUserPrivileges,
   } = userContext();
 
   useEffect(() => {
     (async () => {
-      await verifySuperUser();
+      await setUserPrivileges();
       setIsLoading(false);
     })();
-  }, [verifySuperUser]);
+  }, [setUserPrivileges]);
 
   const showSuperUserOptions = () => {
     return (
--- a/src/components/user-profile/index.js
+++ b/src/components/user-profile/index.js
@@ -9,15 +9,15 @@ const UserProfile = () => {
     isSuperUser,
     isSuperUserMode,
     setIsSuperUserMode,
-    makeApiCallToVerifySuperUser,
+    verifySuperUser,
   } = userContext();
 
   useEffect(() => {
     (async () => {
-      await makeApiCallToVerifySuperUser();
+      await verifySuperUser();
       setIsLoading(false);
     })();
-  }, [makeApiCallToVerifySuperUser]);
+  }, [verifySuperUser]);
 
   const showSuperUserOptions = () => {
     return (
--- a/src/components/user-profile/index.js
+++ b/src/components/user-profile/index.js
@@ -1,32 +1,23 @@
 import { useState, useEffect } from 'react';
 import { userContext } from '@store/user/user-context';
 import classNames from '@components/user-profile/user-profile.module.scss';
-import { getUserSelf } from '../../helper-functions/action-handlers';
 
 const UserProfile = () => {
   const [isLoading, setIsLoading] = useState(true);
   const {
     user,
-    setUser,
     isSuperUser,
-    setIsSuperUser,
     isSuperUserMode,
     setIsSuperUserMode,
-    setApiCalledToVerifySuperUser,
+    makeApiCallToVerifySuperUser,
   } = userContext();
 
   useEffect(() => {
-    async function getUserProfile() {
-      const { data } = await getUserSelf();
+    (async () => {
+      await makeApiCallToVerifySuperUser();
       setIsLoading(false);
-      setUser(data);
-      if (data) {
-        setIsSuperUser(Boolean(data.roles?.super_user));
-        setApiCalledToVerifySuperUser(true);
-      }
-    }
-    getUserProfile();
-  }, [setUser, setIsSuperUser, setApiCalledToVerifySuperUser]);
+    })();
+  }, [makeApiCallToVerifySuperUser]);
 
   const showSuperUserOptions = () => {
     return (
--- a/src/pages/[id]/index.js
+++ b/src/pages/[id]/index.js
@@ -65,7 +65,6 @@ const MemberProfile = ({ imageLink, user, contributions, errorMessage }) => {
   const { first_name = '', last_name = '' } = user;
   const memberName = `${first_name} ${last_name} | Member Real Dev Squad`;
 
-  // const { query } = useRouter(); // this needs to be changed
   const devUser = !!router.query.dev;
 
   return (
--- a/src/pages/[id]/index.js
+++ b/src/pages/[id]/index.js
@@ -20,19 +20,19 @@ import {
   HEIGHT_40PX,
 } from '@constants/profile-image';
 import { useEffect, useState, useCallback } from 'react';
-import { getUserSelf } from '@helper-functions/action-handlers';
+import { userContext } from '@store/user/user-context';
 
 const MemberProfile = ({ imageLink, user, contributions, errorMessage }) => {
-  const [isSuperUser, setIsSuperUser] = useState(false);
+  const {
+    isSuperUser,
+    apiCalledToVerifySuperUser,
+    makeApiCallToVerifySuperUser,
+  } = userContext();
   const [activeTasksData, setActiveTasksData] = useState([]);
   const router = useRouter();
   const { id } = router.query;
 
   const fillActiveTasksArray = useCallback(async () => {
-    const { data } = await getUserSelf();
-    if (data) {
-      setIsSuperUser(Boolean(data.roles?.super_user));
-    }
     const tasksURL = getActiveTasksURL(id);
     try {
       const tasksResponse = await fetch(tasksURL);
@@ -51,6 +51,11 @@ const MemberProfile = ({ imageLink, user, contributions, errorMessage }) => {
     return <NotFound errorMsg={errorMessage} />;
   }
 
+  if (!isSuperUser) {
+    if (!apiCalledToVerifySuperUser) {
+      makeApiCallToVerifySuperUser();
+    }
+  }
   if (!isSuperUser && !user.roles?.member) {
     const errorMsg = `The Member Page for ${
       id.charAt(0).toUpperCase() + id.slice(1)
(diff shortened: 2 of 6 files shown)


harshith-venkatesh:
Why can't we think of making the call if and when it is undefined?
Instead of making call everytime....

heyrandhir:
> Why can't we think of making the call if and when it is undefined? Instead of making call everytime....

Hi @harshith-venkatesh I have now added a check before making an API call. If the API is called previously it doesn't make the call again.

src/pages/[id]/index.js
```diff
@@ -50,11 +47,26 @@ const MemberProfile = ({ imageLink, user, contributions, errorMessage }) => {
...
+      id.charAt(0).toUpperCase() + id.slice(1)
+    } is not yet generated.`;
+    return <NotFound errorMsg={errorMsg} />;
+  }
   const { first_name = '', last_name = '' } = user;
   const memberName = `${first_name} ${last_name} | Member Real Dev Squad`;
 
-  const { query } = useRouter(); // this needs to be changed
-  const devUser = !!query.dev;
+  // const { query } = useRouter(); // this needs to be changed
```
Comment (prakashchoudhary07) [inline src/pages/[id]/index.js:68]: NIT: If not required we should just remove it

src/pages/[id]/index.js
```diff
@@ -50,11 +47,26 @@ const MemberProfile = ({ imageLink, user, contributions, errorMessage }) => {
     fillActiveTasksArray();
   }, [fillActiveTasksArray]);
 
+  if (errorMessage) {
+    return <NotFound errorMsg={errorMessage} />;
+  }
+
+  if (!isSuperUser) {
+    if (!apiCalledToVerifySuperUser) {
+      makeApiCallToVerifySuperUser();
```
Comment (prakashchoudhary07) [inline src/pages/[id]/index.js:56]: Since makeApiCallToVerifySuperUser is async shouldn't there be an await here?

Comment (heyrandhir) [reply src/pages/[id]/index.js:56]: I tried adding await here like this
```
    if (!apiCalledToVerifySuperUser) {
      (async () => await makeApiCallToVerifySuperUser())();
    }
``` 
but I am getting an error in the git log when committing as follows 
```
error  Redundant use of `await` on a return value  no-return-await
```

src/pages/[id]/index.js
Comment (heyrandhir) [reply src/pages/[id]/index.js:68]: sure we can remove this.

prakashchoudhary07 (approved):

src/components/user-profile/index.js
```diff
@@ -1,30 +1,23 @@
...
 const UserProfile = () => {
   const [isLoading, setIsLoading] = useState(true);
   const {
     user,
-    setUser,
     isSuperUser,
-    setIsSuperUser,
     isSuperUserMode,
     setIsSuperUserMode,
+    makeApiCallToVerifySuperUser,
```
ankushdharkar, line 12: Nit: This can be a shorter verb "verifySuperUser". We don't need to write the imperative steps of _how_ we are doing it.
Otherwise, many functions will start looking like this:
- callFunctionFromClassWalletToCheckForAmountRemaining
- checkIfEmailIsValidByMatchingRegularExpressionForEmail

src/components/user-profile/index.js
```diff
@@ -1,30 +1,23 @@
...
       setIsLoading(false);
-      setUser(data);
-      if (data) {
-        setIsSuperUser(data.roles.super_user);
-      }
-    }
-    getUserProfile();
-  }, [setUser, setIsSuperUser]);
+    })();
+  }, [makeApiCallToVerifySuperUser]);
```
ankushdharkar, line 20: API Call in useEffect :O

src/pages/[id]/index.js
```diff
@@ -50,11 +47,25 @@ const MemberProfile = ({ imageLink, user, contributions, errorMessage }) => {
     fillActiveTasksArray();
   }, [fillActiveTasksArray]);
 
+  if (errorMessage) {
```
Comment (ankushdharkar) [inline src/pages/[id]/index.js:50]: Nice!

src/pages/[id]/index.js
```diff
@@ -50,11 +47,25 @@ const MemberProfile = ({ imageLink, user, contributions, errorMessage }) => {
...
+    return <NotFound errorMsg={errorMessage} />;
+  }
+
+  if (!isSuperUser) {
+    if (!apiCalledToVerifySuperUser) {
```
Comment (ankushdharkar) [inline src/pages/[id]/index.js:55]: Nit: Could have been just one If here.

src/pages/[id]/index.js
```diff
@@ -50,11 +47,25 @@ const MemberProfile = ({ imageLink, user, contributions, errorMessage }) => {
...
+      makeApiCallToVerifySuperUser();
+    }
+  }
+  if (!isSuperUser && !user.roles?.member) {
+    const errorMsg = `The Member Page for ${
+      id.charAt(0).toUpperCase() + id.slice(1)
```
Comment (ankushdharkar) [inline src/pages/[id]/index.js:61]: This seems a little out of the ordinary.
Why not use the first and last names here, and show userId as it is?

src/store/user/user-context.js
```diff
@@ -10,6 +11,18 @@ export const UserContextProvider = ({ children }) => {
   const [showMemberRoleUpdateModal, setShowMemberRoleUpdateModal] =
     useState(false);
   const [selectedMember, setSelectedMember] = useState(null);
+  const [apiCalledToVerifySuperUser, setApiCalledToVerifySuperUser] =
```
ankushdharkar, line 14: Could we please use simpler names here, as described here https://github.com/Real-Dev-Squad/website-members/pull/410/files#r952055322?

src/store/user/user-context.js
```diff
@@ -10,6 +11,18 @@ export const UserContextProvider = ({ children }) => {
...
+    useState(false);
+
+  const makeApiCallToVerifySuperUser = async () => {
+    const { data } = await getUserSelf();
```
ankushdharkar, line 18: What about error handling? What happens if the request encounters a 4xx or 5xx?

ankushdharkar (changes requested):

src/store/user/user-context.js
heyrandhir: yeah right. Updated the variable names.

src/pages/[id]/index.js
Comment (heyrandhir) [reply src/pages/[id]/index.js]: Yeah, I have updated the code now to use the first name & last name.

src/pages/[id]/index.js
Comment (heyrandhir) [reply src/pages/[id]/index.js]: merged both the conditions into one. Thanks.

src/components/user-profile/index.js
heyrandhir: Yeah Right. Thanks 🙂

src/pages/[id]/index.js
Comment (heyrandhir) [reply src/pages/[id]/index.js:46]: Thanks 😇

src/components/user-profile/index.js
heyrandhir: I need to read more about this. I will create a new issue and work on this part. Resolving this so that it can be shipped to production and used.

src/store/user/user-context.js
```diff
@@ -10,6 +11,24 @@ export const UserContextProvider = ({ children }) => {
...
+  const verifySuperUser = async () => {
+    setUser(null);
+    setIsSuperUser(false);
+    setSuperUserVerified(false);
+    try {
+      const { data } = await getUserSelf();
+      if (data) {
+        setUser(data);
+        setIsSuperUser(Boolean(data.roles?.super_user));
+        setSuperUserVerified(true);
```
ivinayakg, line 25: if both the `isSuperUser` and `superUserVerified` are getting settled to true at the same time then can't we just switch to one state instead of multiple cause I found it redundant?

heyrandhir: `isSuperUser` checks if a user is or is not a superUser but `superUserVerified` checks if the API call to verify super User is completed. So the next time we don't make an API call again to verify whether or not the user is a superUser.

src/store/user/user-context.js
```diff
@@ -10,6 +11,27 @@ export const UserContextProvider = ({ children }) => {
...
+    let selfSuperUserVerified = false;
+    try {
+      const { data } = await getUserSelf();
+      if (data) {
+        selfUser = data;
+        selfIsSuperUser = Boolean(data.roles?.super_user);
+        selfSuperUserVerified = true;
+      }
+    } catch {
+      // by default handled in the initial local variables
```
ankushdharkar, line 28: Nit: console.error here would have been nice

src/store/user/user-context.js
```diff
@@ -10,6 +11,27 @@ export const UserContextProvider = ({ children }) => {
...
+  const verifySuperUser = async () => {
+    let selfUser = null;
+    let selfIsSuperUser = false;
+    let selfSuperUserVerified = false;
+    try {
+      const { data } = await getUserSelf();
+      if (data) {
+        selfUser = data;
+        selfIsSuperUser = Boolean(data.roles?.super_user);
+        selfSuperUserVerified = true;
```
ankushdharkar, line 25: Why is this needed? Why can't we just reuse `selfIsSuperUser`?

src/components/user-profile/index.js
```diff
@@ -1,30 +1,23 @@
...
       setIsLoading(false);
-      setUser(data);
-      if (data) {
-        setIsSuperUser(data.roles.super_user);
-      }
-    }
-    getUserProfile();
-  }, [setUser, setIsSuperUser]);
+    })();
+  }, [verifySuperUser]);
```
ankushdharkar, line 20: This should ideally not be the responsibility of the component. It should exist in an API util, or data layer.

src/pages/[id]/index.js
```diff
@@ -50,11 +43,21 @@ const MemberProfile = ({ imageLink, user, contributions, errorMessage }) => {
     fillActiveTasksArray();
   }, [fillActiveTasksArray]);
 
+  if (errorMessage) {
+    return <NotFound errorMsg={errorMessage} />;
+  }
+
+  if (!isSuperUser && !superUserVerified) {
```
Comment (ankushdharkar) [inline src/pages/[id]/index.js:50]: `superUserVerified` checks for whether the call was made or not? From the name, it seems like it verifies if the current user is indeed a super user.

src/store/user/user-context.js
```diff
@@ -10,6 +11,27 @@ export const UserContextProvider = ({ children }) => {
...
     useState(false);
   const [selectedMember, setSelectedMember] = useState(null);
+  const [superUserVerified, setSuperUserVerified] = useState(false);
+
+  const verifySuperUser = async () => {
+    let selfUser = null;
+    let selfIsSuperUser = false;
+    let selfSuperUserVerified = false;
+    try {
+      const { data } = await getUserSelf();
```
ankushdharkar, line 21: If the API fails/errors, will we never set `selfSuperUserVerified`, potentially causing infinite loops in consuming components e.g [here](https://github.com/Real-Dev-Squad/website-members/pull/410/files#r953219362)?

ankushdharkar (changes requested):

src/store/user/user-context.js
heyrandhir: probably the name 'selfSuperUserVerified`confused you and vinayak-g. I have changed it to `selfUserApiCalled`. This variable prevents us from making multiple API calls.

src/components/user-profile/index.js
heyrandhir: Agreed. I will create a new ticket to move this into separate API util or data layer.

src/pages/[id]/index.js
Comment (heyrandhir) [reply src/pages/[id]/index.js]: yeah, the name `superUserVerified ` was confusing. renamed it to `userApiCalled`

src/store/user/user-context.js
heyrandhir: even if the API fails/errors after the try-catch block we call `setUserApiCalled` which will set it to `false`( the initial value we initialized `selfUserApiCalled` with).

heyrandhir:
the isOption in the code will be fetched from context. it will be taken care in the refactor PR. https://github.com/Real-Dev-Squad/website-members/issues/420

src/pages/[id]/index.js
```diff
@@ -50,11 +43,21 @@ const MemberProfile = ({ imageLink, user, contributions, errorMessage }) => {
     fillActiveTasksArray();
   }, [fillActiveTasksArray]);
 
+  if (errorMessage) {
+    return <NotFound errorMsg={errorMessage} />;
+  }
+
+  if (!isSuperUser && !userApiCalled) {
+    setUserPrivileges();
+  }
```
Comment (heyrandhir) [inline src/pages/[id]/index.js:52]: This is a potential threat that can cause the page to infinitely re-render. I have created a ticket for this https://github.com/Real-Dev-Squad/website-members/issues/419. Will be refactoring it there.

PR #413, opened 2022-08-18
Make New Users Profile clickable for SuperUsers

### LINKED ISSUE #407: Redirect to Non Members Page upon clicking their Profile from Members Page.
To view the profile page of NonMembers the SuperUser needs to visit members.realdevsquad.com/userName. This should happen at the click of their profile like the members.
Note: This feature should only be enabled for superUsers.

heyrandhir:
### What is the change?
This is a stacked PR & should be merged after PR  https://github.com/Real-Dev-Squad/website-members/pull/410
closes #407

- [x] for superUser make the nonmembers or new members' profile clickable 
- [x] This feature will work once superUser holds the alt key and click on their profile. 

### Is it bug?
No

- Steps to repro
- Expected
- Actual

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [x] Web

#### Browsers

- [x] Chrome
- [ ] Safari
- [ ] Firefox

### DIFF
--- a/src/components/member-card/card.module.scss
+++ b/src/components/member-card/card.module.scss
@@ -14,6 +14,10 @@
   height: 50px;
 }
 
+.removedPointerEvent {
+  pointer-events: none;
+}
+
 .nameOfPerson {
   padding: 10px 10px 0 10px;
   font-size: 1.43rem;
--- a/src/components/member-card/card.module.scss
+++ b/src/components/member-card/card.module.scss
@@ -14,7 +14,7 @@
   height: 50px;
 }
 
-.removedPointerEvent {
+.newUsers {
   pointer-events: none;
 }
 
--- a/src/components/new-members/index.js
+++ b/src/components/new-members/index.js
@@ -5,19 +5,43 @@ import styles from '@components/new-members/new-members.module.scss';
 import { membersContext } from '@store/members/members-context';
 import { searchMemberContext } from '@store/search-members/searchMembers-context';
 import { searchMembers } from '@helper-functions/search-members';
+import Link from 'next/link';
+import { userContext } from '@store/user/user-context';
 
 // returns card which shows details of new member
-const renderNewMember = (newMember, isOptionKey) => (
-  <div className={styles.containerForNewMember}>
-    <Card
-      developerInfo={newMember}
-      isMember={false}
-      isOptionKey={isOptionKey}
-    />
-  </div>
-);
+const renderNewMember = (newMember, isOptionKey) => {
+  return (
+    <div className={styles.containerForNewMember}>
+      <Card
+        developerInfo={newMember}
+        isMember={false}
+        isOptionKey={isOptionKey}
+      />
+    </div>
+  );
+};
+
+const renderNewMemberAsClickable = (newMember, isOptionKey) => {
+  return (
+    <Link
+      prefetch={false}
+      href={{
+        pathname: '/[id]',
+        query: {
+          first_name: `${newMember.first_name || ''}`,
+          last_name: `${newMember.last_name || ''}`,
+        },
+      }}
+      as={`/${newMember.username}`}
+      key={newMember.username}
+    >
+      {renderNewMember(newMember, isOptionKey)}
+    </Link>
+  );
+};
 
 const NewMemberList = ({ isOptionKey }) => {
+  const { isSuperUser } = userContext();
   const {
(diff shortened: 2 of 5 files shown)


prakashchoudhary07 (approved):
LGTM :+1:

src/components/new-members/index.js
```diff
@@ -28,7 +52,9 @@ const NewMemberList = ({ isOptionKey }) => {
       <div className={styles.container}>
         {filterMembers.map((newMember) => (
           <React.Fragment key={newMember.id}>
-            {renderNewMember(newMember, isOptionKey)}
+            {isSuperUser && isOptionKey
+              ? renderNewMemberAsClickable(newMember, isOptionKey)
+              : renderNewMember(newMember, isOptionKey)}
```
heyrandhir, line 57: This logic can be moved inside the renderNewMemberAsClickable

src/components/member-card/card.module.scss
```diff
@@ -14,6 +14,10 @@
   height: 50px;
 }
 
+.removedPointerEvent {
+  pointer-events: none;
+}
+
```
heyrandhir, line 20: we can better the class name here. It's synonymous with the event its doing now.

heyrandhir:
Now we refer to the New Non-Members as New Users.

src/components/new-members/index.js
```diff
@@ -5,19 +5,46 @@ import styles from '@components/new-members/new-members.module.scss';
...
+};
+
+const renderNewUsers = (newMember, isOptionKey, isSuperUser) => {
+  if (isSuperUser && isOptionKey) {
+    return (
+      <Link
+        prefetch={false}
+        href={{
+          pathname: '/[id]',
+          query: {
```
ankushdharkar, line 31: Why is this query needed?

src/components/new-members/index.js
```diff
@@ -28,7 +55,7 @@ const NewMemberList = ({ isOptionKey }) => {
       <div className={styles.container}>
         {filterMembers.map((newMember) => (
           <React.Fragment key={newMember.id}>
-            {renderNewMember(newMember, isOptionKey)}
+            {renderNewUsers(newMember, isOptionKey, isSuperUser)}
```
ankushdharkar, line 58: Why not let the `renderNewUsers` deal with the super user context itself, instead of passing it as an argument?

ankushdharkar (changes requested):

src/components/new-members/index.js
ankushdharkar: Maybe it should handle the Option key within `renderNewUsers` too?

ankushdharkar: Also, I feel it should be called `renderNewUser` (Singular)

src/components/new-members/index.js
heyrandhir: yeah, this is not needed. I will remove it.

src/components/new-members/index.js
heyrandhir: > Why not let the `renderNewUsers` deal with the super user context itself, instead of passing it as an argument?

I thought instead of fetching from the context for every new user. it would be better if I get `isSuperUser` from context once and pass it to each new user card component. Let me know which approach is more optimized. I will change accordingly.

heyrandhir: > Also, I feel it should be called `renderNewUser` (Singular)

yeah Right. My Bad. I will change it.

heyrandhir: > Maybe it should handle the Option key within `renderNewUsers` too?

since we get isOptionKey as props and not from context. We need to pass it to `renderNewUsers`

sumitd94: >I thought instead of fetching from the context for every new user. it would be better if I get isSuperUser from context once and pass it to each new user card component. Let me know which approach is more optimized. I will change accordingly.
Do not pass the argument, let the function handle that please.

heyrandhir: > > I thought instead of fetching from the context for every new user. it would be better if I get isSuperUser from context once and pass it to each new user card component. Let me know which approach is more optimized. I will change accordingly.
> 
> Do not pass the argument, let the function handle that please.

Sure @sumitd94 . May I also know the reason behind this? Tried to google but couldn't get the answer 🥲

PR #421, opened 2022-08-27
added global alt key listener

### LINKED ISSUE #420: Is Option key should be in a global context.
## What is the problem?
- currently, we are handling isOption key for all the components separately, it should be set up in the global context and if the option key is pressed it is set to true for all the components

- A discussion with @Neha is left regarding the `role=button` when that is done there might be more changes to the PR that is raised

vinayak-trivedi:
### What is the change?

closes #420 . Added a listener on the root div which listens for alt key presses and sets isOptionKey to true.

### Is it bug?
no

### \*Dev Tested?

- [X] Yes
- [ ] No

#### Platforms

- [X] Web

#### Browsers

- [X] Chrome
- [ ] Safari
- [ ] Firefox

### DIFF
--- a/src/pages/_app.js
+++ b/src/pages/_app.js
@@ -9,12 +9,19 @@ import { MembersProvider } from '@store';
 import { UserContextProvider } from '@store/user/user-context';
 import { SearchMemberProvider } from '@store/search-members/searchMembers-context';
 import { usePostHog } from 'next-use-posthog';
+import { KeyboardContextProvider } from '@store/keyboard/context';
 
 const MyApp = (props) => {
   const { Component, pageProps } = props;
   const router = useRouter();
 
   const [loading, setLoading] = useState(false);
+  const [isOptionKey, setIsOptionKey] = useState(false);
+
+  const initialvalue = {
+    isOptionKey,
+    setIsOptionKey,
+  };
 
   useEffect(() => {
     const handleStart = (url) => url !== router.asPath && setLoading(true);
@@ -43,12 +50,23 @@ const MyApp = (props) => {
       <UserContextProvider>
         <MembersProvider>
           <SearchMemberProvider>
-            <div className={classNames.root}>
-              <div className={classNames.main}>
-                {loading && <Spinner />}
-                <Component {...pageProps} />
+            <KeyboardContextProvider value={initialvalue}>
+              <div
+                className={classNames.root}
+                role="button"
+                tabIndex={0}
+                onKeyDown={(e) => {
+                  if (e.altKey) {
+                    setIsOptionKey(true);
+                  }
+                }}
+              >
+                <div className={classNames.main}>
+                  {loading && <Spinner />}
+                  <Component {...pageProps} />
+                </div>
               </div>
-            </div>
+            </KeyboardContextProvider>
           </SearchMemberProvider>
         </MembersProvider>
       </UserContextProvider>
--- a/src/pages/_app.js
+++ b/src/pages/_app.js
@@ -16,12 +16,6 @@ const MyApp = (props) => {
   const router = useRouter();
 
   const [loading, setLoading] = useState(false);
-  const [isOptionKey, setIsOptionKey] = useState(false);
-
-  const initialvalue = {
-    isOptionKey,
-    setIsOptionKey,
-  };
 
   useEffect(() => {
     const handleStart = (url) => url !== router.asPath && setLoading(true);
@@ -50,20 +44,8 @@ const MyApp = (props) => {
       <UserContextProvider>
         <MembersProvider>
           <SearchMemberProvider>
-            <KeyboardProvider value={initialvalue}>
-              <div
-                className={classNames.root}
-                role="button"
-                tabIndex={0}
-                onKeyDown={(e) => {
-                  if (e.altKey) {
-                    setIsOptionKey(true);
-                  }
-                }}
-                onKeyUp={() => {
-                  setIsOptionKey(false);
-                }}
-              >
+            <KeyboardProvider>
+              <div className={classNames.root}>
                 <div className={classNames.main}>
                   {loading && <Spinner />}
                   <Component {...pageProps} />
--- a/src/pages/_app.js
+++ b/src/pages/_app.js
@@ -9,7 +9,7 @@ import { MembersProvider } from '@store';
 import { UserContextProvider } from '@store/user/user-context';
 import { SearchMemberProvider } from '@store/search-members/searchMembers-context';
 import { usePostHog } from 'next-use-posthog';
-import { KeyboardContextProvider } from '@store/keyboard/context';
+import { KeyboardProvider } from '@store/keyboard/context';
 
 const MyApp = (props) => {
   const { Component, pageProps } = props;
@@ -50,7 +50,7 @@ const MyApp = (props) => {
       <UserContextProvider>
         <MembersProvider>
           <SearchMemberProvider>
-            <KeyboardContextProvider value={initialvalue}>
+            <KeyboardProvider value={initialvalue}>
               <div
                 className={classNames.root}
                 role="button"
@@ -66,7 +66,7 @@ const MyApp = (props) => {
                   <Component {...pageProps} />
                 </div>
               </div>
-            </KeyboardContextProvider>
+            </KeyboardProvider>
           </SearchMemberProvider>
         </MembersProvider>
       </UserContextProvider>
--- a/src/pages/_app.js
+++ b/src/pages/_app.js
@@ -60,6 +60,9 @@ const MyApp = (props) => {
                     setIsOptionKey(true);
                   }
                 }}
+                onKeyUp={() => {
+                  setIsOptionKey(false);
+                }}
               >
                 <div className={classNames.main}>
                   {loading && <Spinner />}
--- a/src/store/keyboard/context.js
+++ b/src/store/keyboard/context.js
@@ -15,12 +15,14 @@ export const KeyboardProvider = ({ children }) => {
         role="button"
         tabIndex={0}
         onKeyDown={(e) => {
-          if (e.altKey) {
+          if (e.key === 'Alt') {
             setIsOptionKeyPressed(true);
           }
         }}
-        onKeyUp={() => {
-          setIsOptionKeyPressed(false);
+        onKeyUp={(e) => {
+          if (e.key === 'Alt') {
+            setIsOptionKeyPressed(false);
+          }
         }}
       >
         {children}
--- a/src/store/keyboard/context.js
+++ b/src/store/keyboard/context.js
@@ -1,11 +1,31 @@
-import { createContext, useContext } from 'react';
+import { createContext, useContext, useState } from 'react';
 
 const KeyboardContext = createContext();
 
-export const KeyboardProvider = ({ children, value }) => {
+export const KeyboardProvider = ({ children }) => {
+  const [isOptionKey, setIsOptionKey] = useState(false);
+
+  const initialvalue = {
+    isOptionKey,
+    setIsOptionKey,
+  };
+
   return (
-    <KeyboardContext.Provider value={value}>
-      {children}
+    <KeyboardContext.Provider value={initialvalue}>
+      <div
+        role="button"
+        tabIndex={0}
+        onKeyDown={(e) => {
+          if (e.altKey) {
+            setIsOptionKey(true);
+          }
+        }}
+        onKeyUp={() => {
+          setIsOptionKey(false);
+        }}
+      >
+        {children}
+      </div>
     </KeyboardContext.Provider>
   );
 };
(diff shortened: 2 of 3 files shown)


RitikJaiswal75 (commented):
Approved

src/store/keyboard/context.js
```diff
@@ -0,0 +1,19 @@
+import { createContext, useContext } from 'react';
+
+const keyboardContext = createContext();
```
rajat-mehra05, line 3: Can you please capitalize the initial letter of `keyboardContext` here ?

src/store/keyboard/context.js
```diff
@@ -0,0 +1,19 @@
...
+
+export const KeyboardContextProvider = ({ children, value }) => {
```
rajat-mehra05, line 5: I think `KeyboardProvider` would be better naming convention here to avoid confusion since we're already using `keyboardContext.Provider` below

rajat-mehra05 (changes requested):

src/store/keyboard/context.js
vinayak-trivedi: will do that

src/store/keyboard/context.js
vinayak-trivedi: I think it should not be KeyboardProvider, because it is providing the context of that, and keyboardProvider would be more confusing in my opinion

rajat-mehra05: > I think it should not be KeyboardProvider, because it is providing the context of that, and keyboardProvider would be more confusing in my opinion

Please for check for reference: https://kentcdodds.com/blog/how-to-use-react-context-effectively

vinayak-trivedi: thanks for sharing, making the changes

src/pages/_app.js
```diff
@@ -43,12 +50,23 @@ const MyApp = (props) => {
...
+            <KeyboardProvider value={initialvalue}>
+              <div
+                className={classNames.root}
+                role="button"
+                tabIndex={0}
+                onKeyDown={(e) => {
+                  if (e.altKey) {
+                    setIsOptionKey(true);
+                  }
+                }}
```
Mir-SA, line 62: when will this _key_ be set to false?

rajat-mehra05 (commented):
LGTM ! 🚀

src/pages/_app.js
vinayak-trivedi: that we will handle in each component separately

Mir-SA: why so?

vinayak-trivedi: I will do as you suggested in our call

Mir-SA (approved):
LGTM!   :shipit:

src/pages/_app.js
```diff
@@ -9,12 +9,19 @@ import { MembersProvider } from '@store';
...
   const { Component, pageProps } = props;
   const router = useRouter();
 
   const [loading, setLoading] = useState(false);
+  const [isOptionKey, setIsOptionKey] = useState(false);
+
+  const initialvalue = {
+    isOptionKey,
+    setIsOptionKey,
+  };
```
samarpan1738, line 24: Shouldn't this be in KeyboardContext?

vinayak-trivedi: I tried it, we will not be able to access it in our root div then, that is why I have put it in here.

samarpan1738: Which is the root div?

src/pages/_app.js
```diff
@@ -43,12 +50,26 @@ const MyApp = (props) => {
...
+            <KeyboardProvider value={initialvalue}>
+              <div
+                className={classNames.root}
+                role="button"
+                tabIndex={0}
+                onKeyDown={(e) => {
+                  if (e.altKey) {
+                    setIsOptionKey(true);
+                  }
+                }}
```
samarpan1738, line 62: Shouldn't this be in KeyboardContext?

vinayak-trivedi: added it in the latest commit

src/store/keyboard/context.js
```diff
@@ -0,0 +1,39 @@
+import { createContext, useContext, useState } from 'react';
+
+const KeyboardContext = createContext();
+
+export const KeyboardProvider = ({ children }) => {
+  const [isOptionKey, setIsOptionKey] = useState(false);
+
+  const initialvalue = {
+    isOptionKey,
```
samarpan1738, line 9: Can we rename it to `isOptionKeyPressed`?

src/store/keyboard/context.js
```diff
@@ -0,0 +1,39 @@
...
+    setIsOptionKey,
```
samarpan1738, line 10: Do we need to put this in context?

vinayak-trivedi: I do think that we have to put it there, because we need to change it there, correct me if I am understanding it wrong

samarpan1738: > because we need to change it there

Where?

vinayak-trivedi: in the div that we are providing below in the keyboard context

src/store/keyboard/context.js
```diff
@@ -0,0 +1,39 @@
...
+        role="button"
+        tabIndex={0}
+        onKeyDown={(e) => {
+          if (e.altKey) {
+            setIsOptionKey(true);
+          }
+        }}
+        onKeyUp={() => {
+          setIsOptionKey(false);
+        }}
```
samarpan1738, line 25: If we don't check for alt key here I guess it'll setIsOptionKey to false on every key up leading to re-render the components where we have used this context
 
```suggestion
        onKeyUp={() => {
          if (e.altKey)
            setIsOptionKey(false);
        }}
```

vinayak-trivedi: yes correct, changed it

src/store/keyboard/context.js
samarpan1738: I think it'll work. Can you run and verify please?

src/store/keyboard/context.js
```diff
@@ -0,0 +1,40 @@
...
+    <KeyboardContext.Provider value={initialvalue}>
+      <div
+        role="button"
+        tabIndex={0}
+        onKeyDown={(e) => {
+          if (e.altKey) {
+            setIsOptionKey(true);
+          }
+        }}
+        onKeyUp={() => {
```
heyrandhir, line 23: event object `e` needs to be passed as the input parameter.

vinayak-trivedi: thanks, it will be changed in the changed in the next commit

src/store/keyboard/context.js
```diff
@@ -0,0 +1,40 @@
...
+          if (e.altKey)
```
heyrandhir, line 24: `{` can be added at the end of line 24

src/store/keyboard/context.js
samarpan1738: As discussed in vc, e.altKey will always be false in onKeyUp (for any key incl. option key)
If any key is pressed (other than option key) then onKeyUp will set `isOptionKey` to false but it won't lead to re-renders since `isOptionKey` value will already be false

src/store/keyboard/context.js
```diff
@@ -0,0 +1,40 @@
...
+        tabIndex={0}
+        onKeyDown={(e) => {
+          if (e.key === 'Alt') {
+            setIsOptionKeyPressed(true);
+          }
+        }}
+        onKeyUp={(e) => {
+          if (e.key === 'Alt') {
+            setIsOptionKeyPressed(false);
+          }
```
vinayak-trivedi, line 25: I am using `e.key == 'Alt` because e.altkey doesn't change to false on `onKeyPressUp`.

samarpan1738: > change to false

change to true

vinayak-trivedi: edited. thanks

src/store/keyboard/context.js
```diff
@@ -0,0 +1,40 @@
...
+    isOptionKeyPressed,
+  };
+
+  return (
+    <KeyboardContext.Provider value={initialvalue}>
+      <div
+        role="button"
+        tabIndex={0}
+        onKeyDown={(e) => {
+          if (e.key === 'Alt') {
```
samarpan1738, line 18: Can we extract this is a method? Like `isAltKey(e)`

src/store/keyboard/context.js
```diff
@@ -0,0 +1,48 @@
...
+    isOptionKeyPressed,
+  };
+
+  function isOptionKey(e) {
+    if (e.key === 'Alt') {
+      return true;
+    }
+
+    return false;
+  }
```
samarpan1738, line 18: This should be defined outside KeyboardProvider component function(in this file only before line 5)
Otherwise this function will be defined every time the component is rendered.

samarpan1738, line 18:
```suggestion
  function isOptionKey(e) {
    return e.key === 'Alt';
  }
```

vinayak-trivedi: done in the latest commit

samarpan1738 (approved):
LGTM. 
Thanks for making all the requested changes quickly.

src/store/keyboard/context.js
```diff
@@ -0,0 +1,44 @@
...
+const KeyboardContext = createContext();
+
+function isOptionKey(e) {
+  return e.key === 'Alt';
+}
+
+export const KeyboardProvider = ({ children }) => {
+  const [isOptionKeyPressed, setIsOptionKeyPressed] = useState(false);
+
+  const initialvalue = {
```
shubham-y, line 12:
```suggestion
  const initialValue = {
```

src/store/keyboard/context.js
```diff
@@ -0,0 +1,44 @@
...
+      </div>
+    </KeyboardContext.Provider>
+  );
+};
+
+export const useKeyboardContext = () => {
+  const context = useContext(KeyboardContext);
+  if (!context)
+    throw new Error(`keyboardcontext context can only  
+        be used in a component wrapped with keyboardcontext`);
```
shubham-y, line 42:
```suggestion
    throw new Error(`keyboardContext context can only  
        be used in a component wrapped with keyboardContext`);
```

src/store/keyboard/context.js
```diff
@@ -0,0 +1,44 @@
...
+
+export const KeyboardProvider = ({ children }) => {
+  const [isOptionKeyPressed, setIsOptionKeyPressed] = useState(false);
+
+  const initialvalue = {
+    isOptionKeyPressed,
+  };
+
+  return (
+    <KeyboardContext.Provider value={initialvalue}>
```
shubham-y, line 17:
```suggestion
    <KeyboardContext.Provider value={initialValue}>
```

shubham-y (changes requested):
@vinayak-trivedi Now that you have handled the option key press event in context, we should also use the `useKeyboardContext` in all the components requiring option key event like [here](https://github.com/Real-Dev-Squad/website-members/blob/develop/src/components/home/index.js#L21)

src/store/keyboard/context.js
```diff
@@ -0,0 +1,44 @@
...
+      <div
+        role="button"
+        tabIndex={0}
+        onKeyDown={(e) => {
+          if (isOptionKey(e)) {
+            setIsOptionKeyPressed(true);
+          }
+        }}
+        onKeyUp={(e) => {
+          if (isOptionKey(e)) {
```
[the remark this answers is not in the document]
vinayak-trivedi: It is common to name event as 'e' that shouldn't be confusing

shubham-y (approved):
LGTM 👍

src/store/keyboard/context.js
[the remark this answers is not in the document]
Maheima: Got it, Thanks for letting me know about this

Maheima (approved):

src/store/keyboard/context.js
```diff
@@ -0,0 +1,44 @@
...
+  const initialValue = {
+    isOptionKeyPressed,
+  };
+
+  return (
+    <KeyboardContext.Provider value={initialValue}>
+      <div
+        role="button"
+        tabIndex={0}
+        onKeyDown={(e) => {
```
vinayak-trivedi, line 21: there is a discussion pending with neha regarding `role= button`, when it is done I will raise a PR for that if this way of doing it is wrong

ankushdharkar: @vinayak-trivedi Could you please add the ticket that is added for this?

src/store/keyboard/context.js
```diff
@@ -0,0 +1,44 @@
+import { createContext, useContext, useState } from 'react';
+
+const KeyboardContext = createContext();
+
+function isOptionKey(e) {
+  return e.key === 'Alt';
```
ankushdharkar, line 6: Can we fetch this from constants please?

ankushdharkar (changes requested):

prakashchoudhary07 (approved):
LGTM :ship:

PR #426, opened 2022-09-02
fix: removed Math.random() from src of <motion.img/>

### LINKED ISSUE #423: Preventing image re-downloads on routing
When we go to members.realdevsquad.com and click on any member, it re-downloads every member's image (when we just need one image of the member we clicked)
Response got on clicking any particular member
image

YashJain24-chief:
### What is the change?

Previously, Math.random() was concatenated with the image URL query, which served no purpose in my opinion. The downside of concatenating Math.random() is that each time we navigate to a different page and return, a new number is generated and concatenated with the image URL. When react detects that an asset has not yet been downloaded, it sends a request for the new assets, resulting in all images being downloaded repeatedly when navigating back and forth.
Closes #423 

### Is it bug?

- Steps to repro - Visit [https://members.realdevsquad.com/](url), click on any member, and check the network tab.
- Expected - Only the image of clicked member has to be downloaded.
- Actual - All the images are being downloaded irrespective of the clicked member 

**Before Changes - When clicking on a particular member**
image

**After Changes**
Before clicking on a particular member, the total number of requests is 29.
image

After clicking on a particular member, the total number of requests is 30. Which means the image of only the clicked user was downloaded.
image

### DIFF
--- a/src/components/member-card/index.js
+++ b/src/components/member-card/index.js
@@ -47,9 +47,7 @@ const Card = ({ developerInfo, isOptionKey }) => {
       )}
       <motion.img
         layoutId={username}
-        src={
-          isMember ? `${img_url}?${Math.random() * 100}` : '/images/Avatar.png'
-        }
+        src={isMember ? `${img_url}` : '/images/Avatar.png'}
         onError={brokenImageHandler}
         className={
           isMember ? classNames.imgContainer : classNames.imgContainerNewMember


Maheima:
I am unable to reproduce the bug. It is just the data of the member we click on that gets downloaded when I click their profile.
image

The images of all the members are being downloaded multiple times when we are on the homepage of member's site though. Also, there are multiple entries of this 
image

harshith-venkatesh (approved):

Maheima:
> @Maheima Notice the network tab > Img while navigating back and forth. When we click on a particular member all the member images are being downloaded, instead, only the image of clicked member has to be downloaded image

This screenshot is from network tab only

sumitd94 (approved):

ankushdharkar:
Well done! Thanks for the quick turnaround :)

Issue #429, opened 2022-09-08
Super User Functionalities Bugs & Minor Improvements


heyrandhir:
- [ ] Get Rid of Super User Options check box 
- [ ] Promote member functionality not working for Super User
- [ ] Hovering on a user card when it can be clicked should show an appropriate 'cursor'

PR #431, opened 2022-09-08
Remove super user options check box and bug fix

heyrandhir:
### What is the change?
Resolves the issues of the ticket #429 

- [x] Get Rid of Super User Options check box
- [x] Promote member functionality not working for Super User
- [x] Hovering on a user card when it can be clicked should show an appropriate 'cursor'

### Is it bug?
 Yes

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [x] Web

#### Browsers

- [x] Chrome
- [ ] Safari
- [ ] Firefox


prakashchoudhary07 (approved):
LGTM

src/components/home/index.js
```diff
@@ -34,12 +42,11 @@ const Home = () => {
         setIsOptionKey(false);
       }}
     >
-      <UserProfile />
       {dev && <SearchBox />}
       <h1 className={styles.heading}>Designers</h1>
       <Designers />
       <h1 className={styles.heading}>{`${BRAND_NAME} ${MEMBERS_TITLE}`}</h1>
-      {isSuperUserMode && (
+      {isSuperUser && (
```
Maheima, line 49: What is the significance of changing isSuperUserMode to isSuperUser?

heyrandhir: Earlier superUser had to click on a checkbox to enable superUserMode and on enabling we set `isSuperUserMode ` to true. Now we got rid of it so directly using `isSuperUser` instead of `isSuperUserMode `.
image

src/components/new-members/new-members.module.scss
```diff
@@ -6,6 +6,21 @@
...
+  padding: 20px 10px;
+  background: white;
+  max-width: 270px;
+  border-radius: 5%;
+  transition: 0.38s box-shadow ease-out;
+  cursor: none;
+  flex: 0 150px;
+  height: 160px;
+  overflow: hidden;
+  margin: auto;
```
rajat-mehra05, line 21: Is this really required here ? Which other `property` can we use here ?

heyrandhir: yeah, the margin is needed. We need the same CSS that we have used earlier for NewUser.

rajat-mehra05: Okay. But we ought to use `padding` instead of `margin` because more often, adding `margin` leads to breaking off `containers`.

heyrandhir: Ok. I changed to `padding` now instead of `margin`. Thanks.

PR #438, opened 2022-09-26
Feature/SignoutFrontend

singh3069:
### Is it bug?
    No
- Steps to repro
- Expected
- Actual

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [x] Web

#### Browsers

- [x] Chrome
- [ ] Safari
- [ ] Firefox

### Before Changes
 As You can see  in the video that before there is not an option for sign-out when user clicks on  his/her name they are directed to My profile page.

[screen-capture (2).webm](https://user-images.githubusercontent.com/73482745/192236019-795e7e72-6050-400d-bf48-962d0eefa5f1.webm)

### After Changes
As You can see  in the video that after my changes there is a drop down added that opens when user clicks on his/her profile with two options 01. My profile and 02. Sign-out.

[screen-capture (3).webm](https://user-images.githubusercontent.com/73482745/192237175-224c0504-f713-4fe9-bb6e-f164a3ca9da6.webm)

### DIFF
--- a/src/components/UI/dropdown/dropdown.module.scss
+++ b/src/components/UI/dropdown/dropdown.module.scss
@@ -47,6 +47,15 @@
   border-top: 1px solid black;
 }
 
+.signOutWrapper > button {
+  all: unset;
+  cursor: pointer;
+}
+
+.signOutWrapper button:focus-visible {
+  outline: black 5px auto;
+}
+
 .signOut:hover {
   color: #49a82e;
   cursor: pointer;
@@ -78,8 +87,6 @@
   align-items: center;
   justify-content: center;
 }
-
-
 
 .dropDownClose {
   width: 180px;
--- a/src/components/UI/dropdown/dropdown.module.scss
+++ b/src/components/UI/dropdown/dropdown.module.scss
@@ -49,53 +49,32 @@
   display: flex;
   flex-direction: column;
   align-items: center;
-  border: 1px solid #ffffff;
-  border-radius: 10px;
-  width: 200px;
-  box-shadow: 0 10px 25px rgba(0, 0, 0, 0.1);
-  background-color: #1d1283;
+  color: #000000;
 }
 
 .dropDownClose {
-  width: 200px;
+  width: 150px;
   box-shadow: 0 10px 25px rgba(0, 0, 0, 0.1);
-  background-color: #1d1283;
+  background-color: white;
   pointer-events: none;
-  inset: 0;
-  animation: fade-out 500ms forwards;
-  // z-index: 50;
+  transition: height 0.25s ease;
+  height: 0;
   position: absolute;
-  right: 25px;
-  top: 25px;
-  border-radius: 10px;
+  z-index: 50;
+  overflow: hidden;
+  right: 40px;
 }
 
 .dropDownOpen {
-  animation: fade-in 500ms forwards;
-  z-index: 50;
+  width: 180px;
+  box-shadow: 0 10px 25px rgba(0, 0, 0, 0.1);
+  background-color: white;
+  height: 180px;
+  transition: height 0.25s ease;
   position: absolute;
-  right: 25px;
-  top: 70px;
-}
-
-@keyframes fade-in {
-  0% {
-    opacity: 0;
-  }
-
-  100% {
-    opacity: 1;
-  }
-}
-
-@keyframes fade-out {
-  0% {
-    opacity: 1;
-  }
-
-  100% {
-    opacity: 0;
-  }
+  z-index: 50;
+  overflow: hidden;
+  right: 35px;
 }
 
 @media screen and (max-width: 970px) {
--- a/src/components/UI/dropdown/dropdown.module.scss
+++ b/src/components/UI/dropdown/dropdown.module.scss
@@ -16,14 +16,8 @@
 }
 .dropdownHeader {
   margin: 10px;
+  cursor: pointer;
 }
-// .userProfilePic {
-//   width: 32px;
-//   height: 32px;
-//   border-radius: 50%;
-//   display: inline;
-//   vertical-align: middle;
-// }
 
 .userGreetMsg {
   display: inline;
@@ -32,17 +26,30 @@
   color: rgb(255, 255, 255);
 }
 
-.dropdownContent a:hover {
-  color: #49a82e;
-}
-
-.dropdownContent div > p:hover {
-  color: #49a82e;
-}
-
-.dropdownContent div > p {
+.myProfileWrapper {
+  height: 50%;
   text-align: center;
-  padding-bottom: 10px;
+  border-bottom: 1px solid black;
+}
+
+.myProfile:hover {
+  color: #49a82e;
+}
+
+.myProfile {
+  color: #000000;
+  text-decoration: none;
+}
+
+.signOutWrapper {
+  height: 50%;
+  text-align: center;
+  border-top: 1px solid black;
+}
+
+.signOut:hover {
+  color: #49a82e;
+  cursor: pointer;
 }
 
 .dropdownContent {
@@ -50,26 +57,52 @@
   flex-direction: column;
   align-items: center;
   color: #000000;
+  border: 2px solid black;
+  height: 100%;
+  border-radius: 8px;
+  width: 100%;
+  justify-items: center;
 }
 
+.dropDownContentWrapper {
+  width: 100%;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+}
+
+.dropDownContentWrapper div {
+  width: 100%;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+// .dropdownContent div>p {
+//   text-align: center;
+//   // padding-bottom: 10px;
+// }
+
 .dropDownClose {
-  width: 150px;
+  width: 135px;
   box-shadow: 0 10px 25px rgba(0, 0, 0, 0.1);
   background-color: white;
+  border-radius: 8px;
   pointer-events: none;
   transition: height 0.25s ease;
   height: 0;
   position: absolute;
   z-index: 50;
   overflow: hidden;
-  right: 40px;
+  right: 35px;
 }
 
 .dropDownOpen {
-  width: 180px;
+  width: 135px;
   box-shadow: 0 10px 25px rgba(0, 0, 0, 0.1);
   background-color: white;
-  height: 180px;
+  border-radius: 8px;
+  height: 115px;
   transition: height 0.25s ease;
   position: absolute;
   z-index: 50;
--- a/src/components/UI/dropdown/dropdown.module.scss
+++ b/src/components/UI/dropdown/dropdown.module.scss
@@ -1,0 +1,3 @@
+.wrapper {
+  background-color: #dcdcdc;
+}
--- a/src/components/UI/dropdown/dropdown.module.scss
+++ b/src/components/UI/dropdown/dropdown.module.scss
@@ -1,3 +1,105 @@
-.wrapper {
-  background-color: #dcdcdc;
+.userGreet {
+  color: #ffffff;
+  margin: 23px;
+}
+
+.userGreet a {
+  padding: 0px;
+}
+
+.userProfilePic {
+  width: 32px;
+  height: 32px;
+  border-radius: 50%;
+  display: inline;
+  vertical-align: middle;
+}
+.dropdownHeader {
+  margin: 10px;
+}
+// .userProfilePic {
+//   width: 32px;
+//   height: 32px;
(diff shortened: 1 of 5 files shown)


src/components/UI/dropdown/dropdown.module.scss
```diff
@@ -0,0 +1,136 @@
...
+
+.myProfileWrapper {
+  padding-bottom: 15px;
+  text-align: center;
+  border-bottom: 1px solid #000000;
+}
+
+.myProfile:hover {
+  color: #49a82e;
+}
```
vinayak-trivedi, line 41: can we put these colors in css variables please?

singh3069: made the changes as you suggested please have look.

src/components/UI/dropdown/dropdown.module.scss
```diff
@@ -0,0 +1,143 @@
...
+.dropdownHeader {
+  margin: 10px;
+  cursor: pointer;
+}
+
+.userGreetMsg {
+  display: inline;
+  vertical-align: middle;
+  margin-right: 10px;
+  color: rgb(255, 255, 255);
```
vinayak-trivedi, line 33: why are we using RGB here instead of hex code?

src/components/UI/dropdown/dropdown.module.scss
```diff
@@ -0,0 +1,143 @@
...
+  border-top: 1px solid $border-color;
+}
+
+.signOutWrapper > button {
+  all: unset;
+  cursor: pointer;
+}
+
+.signOutWrapper button:focus-visible {
+  outline: var(--black) 5px auto;
```
vinayak-trivedi, line 67: why are you using `var(--black)` here, I think this is not valid

src/components/UI/dropdown/dropdown.module.scss
```diff
@@ -0,0 +1,143 @@
...
+}
+
+.dropDown {
+  background-color: $background-color;
+  width: 180px;
+  position: absolute;
+  overflow: hidden;
+  right: 35px;
+  border-radius: 8px;
+  box-shadow: 0 10px 25px rgba(0, 0, 0, 0.1);
```
vinayak-trivedi, line 109: same as above, why are we using RGB here?

src/components/UI/navbar/index.js
```diff
@@ -187,26 +176,13 @@ const Navbar = () => {
...
-                      </a>
-                    </Link>
-                  </div>
+                  <Dropdown
+                    isLoggedIn={isLoggedIn}
+                    USER_PROFILE_URL={USER_PROFILE_URL}
+                    userData={userData}
+                    DEFAULT_AVATAR={DEFAULT_AVATAR}
+                    SIGN_OUT_URL={SIGN_OUT_URL}
+                  />
```
vinayak-trivedi, line 185: why do we have two dropdown components here, I saw it above as well

src/components/UI/dropdown/dropdown.module.scss
singh3069: okay I will make a changed to it

src/components/UI/dropdown/index.js
```diff
@@ -0,0 +1,85 @@
...
+  });
+
+  return (
+    <div className={styles.userGreet} ref={modalClose}>
+      <div
+        role="button"
+        tabIndex={0}
+        className={styles.dropdownHeader}
+        onClick={toggleDropdown}
+        onKeyDown={toggleDropdown}
```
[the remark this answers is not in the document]
singh3069: Because of this error.
image

I found this solution on google and used in our code 
https://stackoverflow.com/questions/48575674/how-to-add-a-keyboard-listener-to-my-onclick-handler

src/components/UI/navbar/index.js
singh3069: Earlier there were two div elements are shown in the navbar file on is for web view and another for mobile view so I made the sperate component and called it twice.

src/components/UI/dropdown/dropdown.module.scss
singh3069: I didn't  added  it was there only I just copied it when I was making the sperate compoonent for drop down please let me know what can I do  or how can I change it .

src/components/UI/dropdown/dropdown.module.scss
singh3069: okay thanks I will changed that.

src/components/UI/dropdown/dropdown.module.scss
singh3069: In the previous code rgb was used many times, so from taking Idea fom that I added the RGB, please let me know what changes can I do.

src/components/UI/dropdown/dropdown.module.scss
Maheima: Even if you are copying it, it is a part of the change in your PR. So, it would be a good idea to write its hex code/variable name here

src/components/UI/dropdown/index.js
```diff
@@ -0,0 +1,85 @@
+import React, { useEffect, useState, useRef } from 'react';
+import Link from 'next/link';
+
```
Maheima, line 3: This space can be removed and similiarly in other files as well

Maheima (changes requested):

src/components/UI/dropdown/dropdown.module.scss
singh3069: Yess,I totally understood what you said Maheima, I will make the required changes.

src/components/UI/dropdown/index.js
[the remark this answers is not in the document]
singh3069: okay,removed that

singh3069: okay removed that also.

Maheima (approved):
LGTM!

src/components/UI/dropdown/index.js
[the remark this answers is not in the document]
vinayak-trivedi: okay got it

src/components/UI/navbar/index.js
vinayak-trivedi: why we are using two navbars for different views?

RitikJaiswal75 (commented):
Tag this branch with the issue.

src/components/UI/dropdown/dropdown.module.scss
```diff
@@ -0,0 +1,143 @@
...
+  }
+}
+
+@media screen and (max-width: 500px) {
+  .dropDown {
+    width: 110px;
+    right: 15px;
+    top: 45px;
+  }
+}
```
sumitd94, line 143: why are we not using `rem, em` here?

src/components/UI/dropdown/index.js
```diff
@@ -0,0 +1,84 @@
+import React, { useEffect, useState, useRef } from 'react';
+import Link from 'next/link';
+import styles from './dropdown.module.scss';
+
+const Dropdown = ({
+  isLoggedIn,
+  USER_PROFILE_URL,
+  SIGN_OUT_URL,
+  userData,
+  DEFAULT_AVATAR,
```
sumitd94, line 10: why are some props upper case and some camelCase?

src/components/UI/dropdown/index.js
```diff
@@ -0,0 +1,84 @@
+import React, { useEffect, useState, useRef } from 'react';
+import Link from 'next/link';
+import styles from './dropdown.module.scss';
+
+const Dropdown = ({
+  isLoggedIn,
+  USER_PROFILE_URL,
+  SIGN_OUT_URL,
```
sumitd94, line 8: should be coming directly from the constants file, why are we passing it as a prop?

src/components/UI/dropdown/index.js
```diff
@@ -0,0 +1,84 @@
+import React, { useEffect, useState, useRef } from 'react';
+import Link from 'next/link';
+import styles from './dropdown.module.scss';
+
+const Dropdown = ({
```
sumitd94, line 5: where are the PropTypes defined for this component?

src/components/UI/dropdown/index.js
```diff
@@ -0,0 +1,84 @@
...
+  isLoggedIn,
+  USER_PROFILE_URL,
+  SIGN_OUT_URL,
+  userData,
+  DEFAULT_AVATAR,
+}) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const toggleDropdown = () => setIsOpen(!isOpen);
```
sumitd94, line 13:
```javascript

const toggleDropdown = () => setIsOpen((prevState) => !prevState);
```

NIT

It's important to perform the state update based on the previous state since react does a batch of state updates.
Very unlikely in our use case, but better to know about it.

src/components/UI/dropdown/index.js
```diff
@@ -0,0 +1,84 @@
...
+  const modalClose = useRef();
+
+  const signOut = () => {
+    fetch(SIGN_OUT_URL, {
+      method: 'GET',
+      credentials: 'include',
+    }).then(() => {
+      window.location.reload();
+    });
+  };
```
sumitd94, line 23: What if it fails? where are we handling the failure?

src/components/UI/dropdown/index.js
```diff
@@ -0,0 +1,84 @@
...
+              </button>
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+};
+
+export default Dropdown;
```
sumitd94, line 84: Cant, we look to create this as a generic re-usable dropdown component? so it can be used across members? Right now it's only used for Signing users out.

sumitd94: Also not sure why is this named a `dropdown` component.
I cannot see any select attribute 🤔

src/components/UI/dropdown/dropdown.module.scss
singh3069: Beacuse I never worked with them and thought px will do the work, can you please let me know why you are suggesting me to use `rem,em` or I can use `%`

src/components/UI/dropdown/index.js
singh3069: because  both props that are  used in camelCase are useState hooks. let me know If I can change it or not.

src/components/UI/dropdown/index.js
singh3069: Because I tried to make the component re-usable.

src/components/UI/dropdown/index.js
singh3069: I guess because This is JavaScript file not a TypeScript file,I  don't know that we have to define PropTypes here also, please guide me.

src/components/UI/dropdown/index.js
singh3069: Thank you for reminding me and I made the changes.

src/components/UI/dropdown/index.js
singh3069: yess got it I will look into it. thank you.

src/components/UI/dropdown/index.js
singh3069: I have created a re-usable dropdown component please let me know if I made any wrong choice.
I don't name this component based on any attribute I basically named it because how it will appear in the UI as you can see it the above video, please let me know if I can change the component  name from dropdown  to any better name.

src/components/UI/dropdown/dropdown.module.scss
Maheima: https://dev.to/theodorusclarence/back-to-basic-should-we-use-rem-em-or-pixel-1hd0

src/components/UI/navbar/index.js
```diff
@@ -112,26 +114,13 @@ const Navbar = () => {
...
-                            ? `${userData.profilePicture}`
-                            : `${DEFAULT_AVATAR}`
-                        }
-                        alt="Profile pic"
-                      />
-                    </a>
-                  </Link>
-                </div>
+                <Dropdown
+                  isLoggedIn={isLoggedIn}
```
Maheima, line 118: When will the value be `false` in the isLoggedIn that you are passing in Dropdown component?
Because what I see is that it is displayed only when isLoggedIn is true.

Also, there is no logical sense in importing these constant values in Navbar component and passing them in Dropdown as props, you should rather import them in Dropdown component itself.

Maheima (changes requested):

src/components/UI/dropdown/index.js
Maheima: I don't think so

PR #454, opened 2022-10-13
Removed `dev=true` flag.

### LINKED ISSUE #449: remove `dev=true`
We should remove `dev=true` feature flag for following features: 
1. promoting user
2. changing task type

fabcodingzest:
### What is the change?

`dev=true` flag removed from contribution task and member card component. A super user will be able to see the setting option with Alt key press without the dev query. 
closes #449

### Is it bug?

No

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [x] Web

#### Browsers
- [x] Chrome
- [ ] Safari
- [ ] Firefox

A super user should be able to test it.

### DIFF
--- a/src/test/unit/components/member-card/index.test.js
+++ b/src/test/unit/components/member-card/index.test.js
@@ -1,5 +1,6 @@
 import { render, screen } from '@testing-library/react';
 import MemberCard from '@components/member-card';
+import { UserContextProvider } from '@store/user/user-context';
 
 describe('Members Card', () => {
   it('should render members full name', () => {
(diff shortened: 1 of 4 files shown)


YashJain24-chief (approved):
Looks good to me!

vinayak-trivedi (approved):
Thanks for the changes @fabcodingzest 
LGTM 👍

src/test/unit/components/member-card/index.test.js
```diff
@@ -11,7 +12,11 @@ describe('Members Card', () => {
       isMember: true,
     };
 
-    render(<MemberCard developerInfo={developerInfo} />);
+    render(
+      <UserContextProvider>
+        <MemberCard developerInfo={developerInfo} />
+      </UserContextProvider>
+    );
```
prakashchoudhary07, line 19: Why do we need UserContext here? are we not getting it from root?

prakashchoudhary07: Wanted to know, where is it being used in the above changes.

prakashchoudhary07: Got it, thanks for the clerification

rupal-2505 (commented):
Looks good to me!

Issue #450, opened 2022-10-13
show message that user is not a member


vinayak-trivedi:
Pages for non members are also generated but they are only accessible to superuser, we should show a message below the user profile picture that this user is not a member

Manas-E:
@vinayak-trivedi If it's not picked yet, I can have a look at it

vinayak-trivedi:
assigned to you @Manas-E

PR #468, opened 2022-10-29
Fix/add member status

Manas-E:
### What is the change?

Created a member status message to notify the user about the status of developer, Issue number #450 

### Is it bug?
No

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [x] Web

#### Browsers

- [x] Chrome
- [ ] Safari
- [x] Firefox


jest.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   moduleNameMapper: {
     '\\.(css|scss|sass|less)$': 'identity-obj-proxy',
     '^@components(.*)$': '<rootDir>/src/components$1',
+    '^@constants(.*)$': '<rootDir>/src/constants$1',
```
vinayak-trivedi, line 6: the `develop` branch already contains this

src/components/member-profile/member-profile.module.scss
```diff
@@ -211,6 +211,10 @@
   .memberDetails {
     text-align: center;
   }
+
+  .memberStatus {
+    justify-content: center !important;
```
vinayak-trivedi, line 216: why do we need to put !important here?

src/components/member-profile/member-profile.module.scss
```diff
@@ -239,3 +243,14 @@
...
   display: none;
 }
+
+.memberStatus {
+  display: flex;
+  margin: 10px 0px;
+  justify-content: flex-start;
+  gap: 5px;
+  color: #576871;
+  font-size: 16px;
```
vinayak-trivedi, line 253: please put all the colors in scss variables

src/components/member-profile/index.js
```diff
@@ -372,6 +372,28 @@ const Profile = (props) => {
...
+              <div className={classNames.memberStatus}>
+                <img
+                  alt="info icon"
+                  src="icons/info.png"
+                  width={20}
+                  height={20}
+                />
+                <span>User is not a Member</span>
+              </div>
+            )}
```
vinayak-trivedi, line 395: we should only show this to only superuser

vinayak-trivedi (changes requested):
Thanks for the PR @Manas-E, suggesting some changes please have a look

jest.config.js
Manas-E: ok it was throwing error during my tests so I added there, its there so i will remove it then

src/components/member-profile/member-profile.module.scss
Manas-E: Actually in desktop it is flex-start so to override it I've put `important` otherwise this property won't work

src/components/member-profile/member-profile.module.scss
Manas-E: Done in the latest commit

src/components/member-profile/index.js
Manas-E: Added Super user condition there

vinayak-trivedi:
can you add ss of the changes

Manas-E:
> can you add ss of the changes

1. For super user condition:  

image

image


2. Added CSS var 

image

3.  Tests 

image

src/components/member-profile/member-profile.module.scss
Manas-E: I have removed the !important tag and modified CSS

src/components/member-profile/index.js
```diff
@@ -372,6 +373,29 @@ const Profile = (props) => {
...
+                <div className={classNames.memberStatus}>
+                  <img
+                    alt="info icon"
+                    src="icons/info.png"
+                    width={20}
+                    height={20}
+                  />
+                  <span>User is not a Member</span>
+                </div>
+              ))}
```
vinayak-trivedi, line 397: rather than repeating the same code, let's create a message variable which will set that message based on if the user is a member or not

vinayak-trivedi (changes requested):

src/components/member-profile/index.js
Manas-E: oh I missed that, here's how it looks now 
image

vinayak-trivedi (approved):
LGTM

rupal-2505 (approved):
Looks good to me!

Manas-E:
Before: 

image


After: 

image

Maheima (approved):
LGTM

PR #470, opened 2022-11-01
fix: cursor is pointer when superuser hover over users

prankurnov15:
### What is the change?
Added the cursor pointer when the superuser hovers on new user profiles.

Changed the two properties : cursor events and cursor pointer in SCSS file.

### Is it bug?
No

- Steps to repro
- Expected
- Actual

### \*Dev Tested?

- [X ] Yes
- [ ] No

#### Platforms

- [X] Web

#### Browsers

- [X] Chrome
- [ ] Safari
- [ ] Firefox

> For visual or interaction changes. Can be video/screenshot.

### DIFF
--- a/src/components/new-members/new-members.module.scss
+++ b/src/components/new-members/new-members.module.scss
@@ -7,14 +7,14 @@
 }
 
 .newUser {
-  pointer-events: none;
+  pointer-events: auto;
   display: flex;
   padding: 20px 10px;
   background: white;
   max-width: 270px;
   border-radius: 5%;
   transition: 0.38s box-shadow ease-out;
-  cursor: none;
+  cursor: default;
   flex: 0 150px;
   height: 160px;
   overflow: hidden;
@@ -29,7 +29,7 @@
   max-width: 270px;
   border-radius: 5%;
   transition: 0.38s box-shadow ease-out;
-  cursor: default;
+  cursor: pointer;
   flex: 0 150px;
   height: 160px;
   overflow: hidden;


Maheima:
If this condition is just for super user then, why do other user able to see the pointer icon too?
Also, in the description of the PR , These are pointers for you to fill in when sending a PR. If there is nothing to mention in them, remove them. 
image

prankurnov15:
@Maheima I have fixed the issues which you pointed out yesterday , you can review it.

src/components/new-members/new-members.module.scss
```diff
@@ -34,9 +34,13 @@
...
   margin: auto;
-  border: 0;
+  border: none;
   box-shadow: none;
   &:hover {
     box-shadow: none;
   }
 }
+.containerForNewMemberHover {
+  @extend .containerForNewMember;
```
harshith-venkatesh, line 44: Impressive 🙌
Good use of @extend

src/components/new-members/new-members.module.scss
```diff
@@ -7,7 +7,7 @@
 }
 
 .newUser {
-  pointer-events: none;
+  pointer-events: auto;
```
harshith-venkatesh, line 10: Isn't the expectation, only super user should perform some action on pointer event and rest shouldn't?
Clarify in case I am thinking wrong here

prankurnov15: Yes any with super user role should be able to have cursor pointer property for rest of the user it should be default cursor pointer.

src/components/new-members/new-members.module.scss
prankurnov15: Thanks for comment.

src/components/new-members/index.js
```diff
@@ -9,8 +9,15 @@ import { userContext } from '@store/user/user-context';
 
 // returns card which shows details of new member
 const renderNewUserCard = (newMember) => {
+  const { isSuperUser } = userContext();
   return (
-    <div className={styles.containerForNewMember}>
+    <div
+      className={
+        isSuperUser
+          ? styles.containerForNewMemberHover
```
Maheima, line 17: Can you please think of better classnames here? Rest everything looks great!

Maheima (changes requested):

src/components/new-members/index.js
prankur34: Sure, I am changing it to something more meanigful.

harshith-venkatesh (approved):

Maheima (approved):
LGTM! Just resolve all the conversations

src/components/new-members/new-members.module.scss
```diff
@@ -34,9 +34,13 @@
   height: 160px;
   overflow: hidden;
   margin: auto;
-  border: 0;
+  border: none;
```
ankushdharkar, line 37: Why was this change made?

ankushdharkar (changes requested):

src/components/new-members/new-members.module.scss
prankur34: Both are valid but shouldn't have changed , I have reverted back the `border: 0`

PR #481, opened 2022-11-04
show relevant options for buttons

muskanbatra-dev:
**What is the change?**
when the super clicks on the setting button to update the role of an existing member he should see only the archive member option.
when he clicks on non-member he should see promote to member and archive button.

**Is it a bug?**
No

***Dev** **Tested?**
 

- [x] Yes

 No

***Tested** **on:**
**Platforms**

- [x] Web

**Browsers**
 

- [x] Chrome

 

- [ ] Safari

 

- [ ] Firefox

**Before Change Screenshots**
Before:
https://www.loom.com/share/cd25966dbd9943da866b34d8efe85b94

After:
https://www.loom.com/share/2a58206fc7ba452aab4aa2e2822db5de

### DIFF
--- a/src/components/member-role-update/index.js
+++ b/src/components/member-role-update/index.js
@@ -14,6 +14,7 @@ const MemberRoleUpdate = () => {
     showMemberRoleUpdateModal,
     setShowMemberRoleUpdateModal,
     selectedMember,
+    isUserMember,
   } = userContext();
 
   const [isUpdating, setIsUpdating] = useState(false);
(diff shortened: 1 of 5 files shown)


src/components/member-role-update/index.js
```diff
@@ -41,6 +42,22 @@ const MemberRoleUpdate = () => {
...
+            type="button"
+            onClick={() => archiveTheMember(selectedMember)}
+          >
+            Archive Member
+          </button>
+          <br />
+          <p>{updateStatus}</p>
+        </>
+      );
+    }
```
vinayak-trivedi, line 59: we need only one button when the user is a member, so instead of repeating yourself, you can put a condition below where the promote to member button only shows up if the user is not a member, also can you change the text in the button to `archive user` if the user is not a member

vinayak-trivedi (changes requested):
Thanks for the PR @muskanbatra-dev 
Requesting one change, and also please add test cases for the changes

PR #503, opened 2023-03-18
Create component to assign users tags

### LINKED ISSUE #502: [Superuser only] Add skill and level tags to every user.
# Issue
A superuser should be able to add skill and level tags on every user.

**Description**
As of now, there is an issue open for the same but just for members. The feature will now be extended to users but it won't show up till they have a members page. Though a superuser should be able to add the skills and levels for non-members as well. this will make it easier when they get their member's page active.

**Tech stack**
- `Next Js`
- `React`
- `js`

RitikJaiswal75:
### What is the change?

Created a component that provides an options to add tags to the user.
closes #502 

### Is it bug?
No

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [x] Web

#### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge

### Screenshot after changes
image

### DIFF
--- a/src/components/member-role-update/index.js
+++ b/src/components/member-role-update/index.js
@@ -3,6 +3,9 @@ import ReactDOM from 'react-dom';
 import Modal from '@components/UI/modal/index';
 import { userContext } from '@store/user/user-context';
 import Spinner from '@components/UI/spinner';
+import MemberTagAssign from '@components/member-tag-assign';
+import { BASE_API_URL } from '@constants/AppConstants';
+import useFetch from '@custom-hooks/useFetch';
 import classNames from './member-role-update.module.scss';
 import {
   archiveMember,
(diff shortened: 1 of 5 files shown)


yesyash (approved):

vinayak-trivedi:
Thanks for the PR @RitikJaiswal75, just had some suggestion can you add a loading state in the UI, when the levels and tags are loading, also can we just improve the UI a bit?

vinayak-trivedi:
> Thanks for the PR @RitikJaiswal75, just had some suggestion can you add a loading state in the UI, when the levels and tags are loading, also can we just improve the UI a bit?

vinayak-trivedi (approved):
LGTM, as discussed other UI improvements will be done in another PR

PR #511, opened 2023-04-05
userpage is not opening without alt+key

Pratiyushkumar:
### What is the change?

### Is it bug?

- Steps to repro
- Expected
- Actual

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [ ] Web

#### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox

### DIFF
--- a/src/components/new-members/index.js
+++ b/src/components/new-members/index.js
@@ -17,14 +17,14 @@ const renderNewUserCard = (newMember) => {
   );
 };
 
-const renderNewUser = (newMember, isSuperUser, handleNewMemberDetailsPage) => {
+const renderNewUser = (newMember, isSuperUser, handleNewMemberClick) => {
   if (isSuperUser) {
     return (
       <div
         role="button"
         tabIndex={-1}
         key={newMember.username}
-        onClick={() => handleNewMemberDetailsPage(newMember.username)}
+        onClick={() => handleNewMemberClick(newMember.username)}
         aria-hidden="true"
       >
         {renderNewUserCard(newMember)}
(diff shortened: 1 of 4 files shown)


src/components/new-members/index.js
```diff
@@ -16,19 +17,18 @@ const renderNewUserCard = (newMember) => {
   );
 };
 
-const renderNewUser = (newMember, isSuperUser) => {
+const renderNewUser = (newMember, isSuperUser, handleNewMemberDetailsPage) => {
```
vinayak-trivedi, line 20: can we name it handleNewMemberClick?

src/components/new-members/index.js
```diff
@@ -16,19 +17,18 @@ const renderNewUserCard = (newMember) => {
...
-        href={{
-          pathname: '/[id]',
-        }}
-        as={`/${newMember.username}`}
+      <div
+        role="button"
+        tabIndex={-1}
         key={newMember.username}
+        onClick={() => handleNewMemberDetailsPage(newMember.username)}
+        aria-hidden="true"
```
Pratiyushkumar, line 28: role="button": - it is helpful for screen readers to identify it as a button but it will provide no button functionality.
tabIndex={-1} - tabIndex allows developers to make HTML elements focusable and a negative value means that element is not reachable via sequential keyboard navigation.
aria-hidden="true" - Adding aria-hidden="true" to an element removes that element and all of its children from the accessibility tree. This can improve the experience for assistive technology.

src/components/new-members/index.js
Pratiyushkumar: yes we can name it, will name it and push the changes again.

vinayak-trivedi (approved):
LGTM

src/components/new-members/index.js
```diff
@@ -16,19 +17,18 @@ const renderNewUserCard = (newMember) => {
...
     return (
-      <Link
-        prefetch={false}
-        href={{
-          pathname: '/[id]',
-        }}
-        as={`/${newMember.username}`}
+      <div
+        role="button"
+        tabIndex={-1}
```
vvaibhavdesai, line 25: tabIndex needs to be a non negative number for the tab to focus on the element since aria-hidden is set to true

vvaibhavdesai (approved):

PR #519, opened 2023-04-12
opening setting options open user page and going back gives application error 

Pratiyushkumar:
…on error

### What is the change?

After the previous PR, there was an issue coming up where on the alt+click event two events were occurring at that same time - 
    1. super user was able to navigate to the details page
    2. also settings icon was visible 
  so the change made was just to stop the event propagation in showModal function and handleNewMemberClick function.

### Is it a bug?
yes it was a bug

- Steps to repro
- Expected
- Actual

### \*Dev Tested?

- [ ] Yes
- [ ] No

#### Platforms

- [ ] Web

#### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox

> For visual or interaction changes. Can be video/screenshot.

### DIFF
--- a/src/components/member-card/super-user-options/index.js
+++ b/src/components/member-card/super-user-options/index.js
@@ -7,6 +7,7 @@ const SuperUserOptions = ({ showSettings, username }) => {
 
   const showModal = (e) => {
     e.preventDefault();
+    e.stopPropagation();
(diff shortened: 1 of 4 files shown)


Pratiyushkumar:
This was already resolved by Prakash

Issue #488, opened 2022-11-16
Fix the hierarchy Link>a>button


Maheima:
As per the standard [HTML docs](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element), a button element shouldn't be used inside anchor (`a`) element. We are doing this at multiple places e.g., in navbar 
![image](https://user-images.githubusercontent.com/42349041/202074935-a98663c6-99ac-4a5d-9aaa-367551599918.png)

kush-x7:
Can you please assign this task to me?

mayurlalwani:
@vinayak-trivedi  Please assign this issue to me.

vinayak-trivedi:
assigned to you @mayurlalwani

PR #523, opened 2023-04-25
fix hierarchy of button and anchor tag

mayurlalwani:
### What is the change?

The change was to fix the heirarchy of button and anchor tag. So I used div element instead of anchor tag.

### Is it bug?

It is not a bug, as it is not breaking anything. The issue is about refactoring and using HTML standard practices. 

### \*Dev Tested?

- [ ] Yes

#### Platforms

- [ ] Web

#### Browsers

- [ ] Chrome

screenshot
 
Issue: #488


src/components/UI/navbar/index.js
```diff
@@ -100,7 +100,7 @@ const Navbar = () => {
             <div className={styles.navBarLogin}>
               {!isLoggedIn && (
                 <Link href={authUrl}>
-                  <a className={styles.btnLogin}>
+                  <div className={styles.btnLogin}>
```
RitikJaiswal75, line 103: if it is a button why keep it in a div cant we directly keep it inside the button tag itself?

mayurlalwani: I had tried this way, but actually div tag has its own styles inside which there is a button. To keep the styles same, I had to keep the div tag also

RitikJaiswal75: For clickable elements its preferred to either use button or anchor other tags are not preferred.

shubhamsigdar1 (commented):
great work @mayurlalwani...not find anything wrong

RitikJaiswal75 (commented):
LGTM

RitikJaiswal75 (approved):

mayurlalwani:
> great work @mayurlalwani...not find anything wrong

@shubhamsigdar1 Can you please approve this PR so that it can be merged

shubhamsigdar1 (approved):
LGTM

mayurlalwani:
@RitikJaiswal75 Can you merge this PR as 2 members have approved?

RitikJaiswal75 (changes requested):
Please write tests for it

iamitprakash (changes requested):
Please add PR details shared by @RitikJaiswal75  and add two images for before and After

PR #525, opened 2023-05-01
fix: if there are no tasks in any of the category then dropdown will …

anishpawaskar:
…not rotate

### What is the change?

If there are no tasks in particular category dropdown will not rotate

### Is it bug?

- Steps to repro
- Expected
- Actual

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [ ] Web

#### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox

Before -
bug

After - 
fixed (2)

### DIFF
--- a/src/components/member-profile/contribution-type/index.js
+++ b/src/components/member-profile/contribution-type/index.js
@@ -28,9 +28,10 @@ const renderActiveTasks = (tasks) => {
 const ContributionType = (props) => {
   const { fullName, type, imageLink, contributions, devUser, tasks } = props;
   const [showMoreContent, setShowMoreContent] = useState(false);
+  const isContribution = type === 'Noteworthy' || type === 'All';
 
   useEffect(() => {
-    if (type === 'Noteworthy' || type === 'All') {
+    if (isContribution) {
       setShowMoreContent(contributions.length > 0);
       return;
     }
@@ -38,7 +39,7 @@ const ContributionType = (props) => {
   }, [tasks.length, contributions.length]);
 
   const showMoreContentHandler = () => {
-    if (type === 'Noteworthy' || type === 'All') {
+    if (isContribution) {
       if (contributions.length > 0) {
         setShowMoreContent(!showMoreContent);
       }
--- a/src/components/member-profile/contribution-type/index.js
+++ b/src/components/member-profile/contribution-type/index.js
@@ -1,5 +1,5 @@
 /* eslint-disable react/no-array-index-key */
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import classNames from '@components/member-profile/contribution-type/contributions-type.module.scss';
 import Contribution from '@components/member-profile/contribution/';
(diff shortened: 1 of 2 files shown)


src/components/member-profile/index.js
```diff
@@ -35,6 +35,7 @@ const renderContributionsTypes = (
   tasks
 ) => {
   const { noteworthy, all } = contributions;
+
```
shubhamsigdar1, line 38: remove space

shubhamsigdar1 (changes requested):

src/components/member-profile/index.js
anishpawaskar: Okay ill do that

src/components/member-profile/contribution-type/index.js
```diff
@@ -27,11 +27,25 @@ const renderActiveTasks = (tasks) => {
...
+      return;
+    }
+    setShowMoreContent(tasks.length > 0);
+  }, [tasks.length, contributions.length]);
 
   const showMoreContentHandler = () => {
-    setShowMoreContent((prevstate) => !prevstate);
+    if (type === 'Noteworthy' || type === 'All') {
+      if (contributions.length > 0) {
+        setShowMoreContent(!showMoreContent);
```
vvaibhavdesai, line 43: why are we directly setting the state ?

src/components/member-profile/contribution-type/index.js
```diff
@@ -27,11 +27,25 @@ const renderActiveTasks = (tasks) => {
...
+      }
+    }
+    if (tasks.length > 0) {
+      setShowMoreContent(!showMoreContent);
```
vvaibhavdesai, line 47: here as well ?
shouldnt it be either `prev=>!prev` or `false/true` based on whatever the situation is?

src/components/member-profile/contribution-type/index.js
```diff
@@ -27,11 +27,25 @@ const renderActiveTasks = (tasks) => {
...
+    if (type === 'Noteworthy' || type === 'All') {
+      setShowMoreContent(contributions.length > 0);
+      return;
+    }
+    setShowMoreContent(tasks.length > 0);
+  }, [tasks.length, contributions.length]);
 
   const showMoreContentHandler = () => {
-    setShowMoreContent((prevstate) => !prevstate);
+    if (type === 'Noteworthy' || type === 'All') {
```
vinayak-trivedi, line 41: can we create a state called like following and use it everywhere it is needed:
```js
const isContribution = type === 'Noteworthy' || type === 'All'
```

anishpawaskar: Yeah it would be better i will add it

src/components/member-profile/contribution-type/index.js
```diff
@@ -27,11 +27,25 @@ const renderActiveTasks = (tasks) => {
 
 const ContributionType = (props) => {
   const { fullName, type, imageLink, contributions, devUser, tasks } = props;
+  const [showMoreContent, setShowMoreContent] = useState(false);
 
-  const [showMoreContent, setShowMoreContent] = useState(true);
+  useEffect(() => {
+    if (type === 'Noteworthy' || type === 'All') {
```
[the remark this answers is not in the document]
anishpawaskar: I will put it in a state as @vinayak-trivedi mentioned so that we can use it whenever it is needed

vvaibhavdesai (approved):

src/components/member-profile/contribution-type/index.js
vvaibhavdesai: this shouldn't be done but the other methods are not working as expected will need to refactor this file.

vvaibhavdesai (approved):

vinayak-trivedi (approved):
LGTM

PR #536, opened 2023-07-08
Provide super user option on member detail page

### LINKED ISSUE #533: Provide super user option on  member detail page
# Super User option
----------------------------------------
### Overview - 
currently, a superuser can only promote a member or archive  a member from the member's page by clicking on the settings button 

### Requirement - 
The superuser should be able to promote/demote a member or archive/unarchive a member when the superuser presses the alt key, the settings button appears when clicking on the settings button super user gets a modal where he is able to perform above mentioned task and the same modal should work on members page as well.

### LINKED ISSUE #539: Add API  update the role in backend

Pratiyushkumar:
### What is the change?

The superuser should be able to promote/demote a member or archive/unarchive a member when the superuser presses the alt key, the settings button appears when clicking on the settings button super user gets a modal where he can perform above mentioned task and the same modal should work on members page as well.

### Is it a bug?
No

### \*Dev Tested?

- [x] Yes
- [ ] No

> For visual or interaction changes. Can be a video/screenshot.

https://github.com/Real-Dev-Squad/website-members/assets/82165483/98480e94-f5b1-4302-b0bb-f62097b51fef

API is under development so unarchiving and demoting functionality is not built so in UI it's not working but  after the API is built, have created follow-up issue and will handle that in that issue - 

https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/users#patch-usersidtemporarydata 

### Follow-up issue 
#548

### DIFF
--- a/src/components/keyboard-handler/index.js
+++ b/src/components/keyboard-handler/index.js
@@ -3,8 +3,11 @@ import { useKeyboardContext } from '@store/keyboard/context';
 import { ALT_KEY } from '@constants/AppConstants';
 
 function isOptionKey(e) {
-  if (e.key === ALT_KEY) e.preventDefault();
-  return e.key === ALT_KEY;
+  if (e.key === ALT_KEY) {
+    e.preventDefault();
+    return true;
+  }
+  return false;
 }
 
 function KeyboardHandler({ children }) {
--- a/src/components/keyboard-handler/index.js
+++ b/src/components/keyboard-handler/index.js
@@ -3,7 +3,7 @@ import { useKeyboardContext } from '@store/keyboard/context';
 import { ALT_KEY } from '@constants/AppConstants';
 
 function isOptionKey(e) {
-  e.preventDefault();
+  if (e.key === ALT_KEY) e.preventDefault();
   return e.key === ALT_KEY;
 }
 
--- a/src/components/keyboard-handler/index.js
+++ b/src/components/keyboard-handler/index.js
@@ -3,6 +3,7 @@ import { useKeyboardContext } from '@store/keyboard/context';
 import { ALT_KEY } from '@constants/AppConstants';
 
 function isOptionKey(e) {
+  e.preventDefault();
   return e.key === ALT_KEY;
 }
 
--- a/src/components/member-profile/index.js
+++ b/src/components/member-profile/index.js
@@ -1,5 +1,6 @@
 /* eslint-disable react/no-array-index-key */
 import React, { useRef, useState } from 'react';
+import { useRouter } from 'next/router';
 import PropTypes from 'prop-types';
 import SocialMediaIcon from '@components/social-media-icon';
 import getBadges from '@components/member-profile/mock/get-badges';
@@ -14,6 +15,9 @@ import { KEY_ESC, KEY_TAB } from '@constants/AppConstants';
 import MemberTaskUpdate from '@components/member-task-update';
 import { useTaskContext } from '@store/tasks/tasks-context';
 import { userContext } from '@store/user/user-context';
+import { useKeyboardContext } from '@store/keyboard/context';
+import SuperUserOptions from '@components/member-card/super-user-options';
+import MemberRoleUpdate from '@components/member-role-update';
 
 const renderBadgeImages = (badges) =>
   badges.map((badge) => (
@@ -66,25 +70,28 @@ const Profile = (props) => {
   } = props;
   const { membersData } = props;
   const isMember = Boolean(membersData?.roles?.member);
+
   const memberStatusMessage = isMember
     ? 'User is a Member'
     : 'User is not a Member';
-
+  const { showMemberRoleUpdateModal, isSuperUser } = userContext();
   const socialMedia = [
     'twitter_id',
     'github_id',
     'linkedin_id',
     'instagram_id',
   ];
+  const { query } = useRouter() || { query: { dev: false } };
+  const { dev } = query;
 
   const { showMemberTaskUpdateModal } = useTaskContext();
-  const { isSuperUser } = userContext();
   const fullName = `${first_name} ${last_name}`;
   const memberName = fullName.trim() || '--';
   const rdsUserName = `@${username}`;
 
   const badges = getBadges(username);
-
+  const { isOptionKeyPressed } = useKeyboardContext();
+  const [showSettings, setShowSettings] = useState(false);
   const [showModal, setShowModal] = useState(false);
   const { register, handleSubmit, errors } = useForm();
   const submitBtnRef = useRef(null);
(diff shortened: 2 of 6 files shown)


src/components/member-role-update/index.js
```diff
@@ -53,21 +57,41 @@ const MemberRoleUpdate = () => {
...
+          </button>
+        ) : (
+          <button
+            className={classNames.moveToMember}
+            type="button"
+            onClick={() => archiveTheMember(selectedMember)}
+          >
+            Archive Member
+          </button>
+        )}
```
vinayak-trivedi, line 93: can we put these in a contant, and then add it here?

src/components/member-profile/index.js
```diff
@@ -379,15 +391,33 @@ const Profile = (props) => {
...
+                  <img
+                    alt="info icon"
+                    src="icons/info.png"
+                    width={20}
+                    height={20}
+                  />
+                  <span>{memberStatusMessage}</span>
+                  {dev && (
+                    <span style={{ marginLeft: '0.8rem' }}>
+                      {isSuperUser && isOptionKeyPressed && (
```
vinayak-trivedi, line 409: if we are checking isSuperUser above do we need to check it here?

vinayak-trivedi (changes requested):
Thanks for the PR @Pratiyushkumar, suggested few changes please check

src/components/member-profile/index.js
Pratiyushkumar: fixed

src/components/member-role-update/index.js
Pratiyushkumar: fixed

src/components/member-profile/index.js
```diff
@@ -367,6 +374,11 @@ const Profile = (props) => {
           {children}
         </Modal>
       )}
+      {isSuperUser && (
+        <div id="memberRoleUpdateModal">
+          {showMemberRoleUpdateModal && <MemberRoleUpdate />}
+        </div>
+      )}
```
vinayak-trivedi, line 381: shouldn't this be like this:

```js
{isSuperUser && showMemberRoleUpdateModal  && (
    <div id="memberRoleUpdateModal">
      <MemberRoleUpdate />
    </div>
)}
```

src/components/member-profile/index.js
```diff
@@ -378,15 +390,25 @@ const Profile = (props) => {
...
+                onMouseLeave={() => setShowSettings(false)}
+              >
                 <img
                   alt="info icon"
                   src="icons/info.png"
                   width={20}
                   height={20}
                 />
                 <span>{memberStatusMessage}</span>
+                <span style={{ marginLeft: '0.8rem' }}>
```
vinayak-trivedi, line 406: let's create a class for it and not use inline style here

vinayak-trivedi (changes requested):

src/components/member-profile/index.js
Pratiyushkumar: what you are mentioning is correct but can not be done here because `react.createPortal` is not allowing to do so,

src/components/member-role-update/index.js
```diff
@@ -26,48 +23,110 @@ const MemberRoleUpdate = () => {
...
+      archived: true,
+    };
+    const { status } = await memberRoleUpdate(id, role);
+    setIsUpdating(false);
+    if (status === 200) {
+      setUpdateStatus('user archived!');
     } else {
       setUpdateStatus('Some error occured, please contact admin');
     }
   };
```
vinayak-trivedi, line 72: can we not create only two functions that will just toggle these roles, instead of having four function, I think that is possible

src/components/member-role-update/index.js
```diff
@@ -26,48 +23,110 @@ const MemberRoleUpdate = () => {
...
+    </button>
+  ) : (
+    <button
+      className={classNames.moveToMember}
+      type="button"
+      onClick={() => archiveTheMember(userId)}
+    >
+      Archive Member
+    </button>
+  );
```
vinayak-trivedi, line 122: as mentioned above, can we not just check for the status of the user, and then on the basis of that change the payload and CTA text?

src/components/member-profile/index.js
```diff
@@ -378,15 +390,25 @@ const Profile = (props) => {
               className={classNames.profilePic}
               alt={fullName}
             />
-            {isSuperUser && (
-              <div className={classNames.memberStatus}>
+            {dev && isSuperUser && isOptionKeyPressed && (
```
sumitd94, line 393: too many `&&` checks here, can we have this in a variable and use that variable here?

src/components/member-role-update/index.js
```diff
@@ -26,48 +23,110 @@ const MemberRoleUpdate = () => {
     `${BASE_API_URL}/users/${selectedMember}`
   );
 
+  const member = userData?.user.roles.member;
+  const archived = userData?.user.roles.archived;
```
sumitd94, line 27:
```suggestion
  const { member } = userData?.user?.roles;
  const { archived } = userData?.user?.roles;
```

src/components/member-role-update/index.js
```diff
@@ -26,48 +23,110 @@ const MemberRoleUpdate = () => {
...
+    };
+    const { status } = await memberRoleUpdate(id, role);
     setIsUpdating(false);
-    if (status === 204) {
-      setUpdateStatus('User moved to member');
+    if (status === 200) {
+      setUpdateStatus('user promoted to a member');
+    } else {
+      setUpdateStatus('some error occured, please contact admin');
+    }
```
sumitd94, line 43: why not wrap it in a try-catch block?

src/components/keyboard-handler/index.js
```diff
@@ -3,6 +3,7 @@ import { useKeyboardContext } from '@store/keyboard/context';
 import { ALT_KEY } from '@constants/AppConstants';
 
 function isOptionKey(e) {
+  if (e.key === ALT_KEY) e.preventDefault();
   return e.key === ALT_KEY;
```
vinayak-trivedi, line 7: so if we are already checking this condition, why don't we just return true inside the code block only

src/components/member-role-update/index.js
```diff
@@ -26,48 +23,98 @@ const MemberRoleUpdate = () => {
...
+    </button>
+  ) : (
+    <button
+      className={classNames.moveToMember}
+      type="button"
+      onClick={() => promoteDemoteAMember(userId)}
+    >
+      Promote to Member
+    </button>
+  );
```
vinayak-trivedi, line 92: https://github.com/Real-Dev-Squad/website-members/pull/536#discussion_r1275181855 I don't think that this comment is resolved yet, so you created a function but I wanted to have just one button for it and check what text to show and change the payload of the function as well by checking the status only

src/components/member-role-update/index.js
```diff
@@ -26,48 +23,98 @@ const MemberRoleUpdate = () => {
...
+  ) : (
+    <button
+      className={classNames.moveToMember}
+      type="button"
+      onClick={() => archiveUnArchiveTheMember(userId)}
+    >
+      Archive Member
+    </button>
+  );
+
```
vinayak-trivedi, line 111: same as above

vinayak-trivedi (approved):
LGTM

anishpawaskar (approved):
LGTM

Issue #550, opened 2023-08-17
Adding reason in the archiving process


gauravsinhaweb:
### Overview

The purpose of this ticket is to enable authorized users to provide a reason for the archiving process. This implementation will facilitate the retrieval of logs pertaining to `archived-details` displaying a clear explanation of why the authorized user archived a specific user.

### User Stories

Authorized users should be able to write a reason while putting users to archived.

### Requirement
- The input field should be a text box.
- The input box for `reason` should be a required field.
- The button should be disabled if the input field is blank/ empty.

### Acceptance criteria

As an Authorized user, I can type text (reason) for archiving the user.

*ETA: 20th September*

**Before:**
![Screenshot 2023-08-17 080809](https://github.com/Real-Dev-Squad/website-members/assets/75125943/ef28f635-8ae2-42b0-81a0-97f1b70b0511)

PR #557, opened 2023-09-13
added reason text box in archiving process

gauravsinhaweb:
**Issue - #550**
### What is the change?

Added Reason text box in the archiving process

### Is it bug?

- Steps to repro
- Expected
- Actualo

### \*Behind Feature flag?

- [x] Yes
- [ ] No

### \*Dev Tested?

- [x] Yes
- [ ] No

#### Platforms

- [ ] Web

#### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox

https://github.com/Real-Dev-Squad/website-members/assets/75125943/5b90d0b9-ff5a-4d58-a40b-92f1c9455da9

Screenshot 2023-09-14 015427
Screenshot 2023-09-14 015253

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,0 +1,1 @@
+NEXT_PUBLIC_BASE_API_URL=https://api.realdevsquad.com
--- a/.env.development
+++ b/.env.development
@@ -1,1 +1,1 @@
-NEXT_PUBLIC_BASE_API_URL=https://staging-api.realdevsquad.com
+NEXT_PUBLIC_BASE_API_URL=https://api.realdevsquad.com
--- a/.env.development
+++ b/.env.development
@@ -1,1 +1,1 @@
-NEXT_PUBLIC_BASE_API_URL=https://api.realdevsquad.com
+NEXT_PUBLIC_BASE_API_URL=https://staging-api.realdevsquad.com
--- a/.env.development
+++ b/.env.development
@@ -1,1 +1,0 @@
-NEXT_PUBLIC_BASE_API_URL=https://api.realdevsquad.com
--- a/src/components/member-role-update/index.js
+++ b/src/components/member-role-update/index.js
@@ -13,7 +13,7 @@ import { memberRoleUpdate } from '../../helper-functions/action-handlers';
 const MemberRoleUpdate = () => {
   const { query } = useRouter() || { query: { dev: false } };
   const { dev } = query;
-  const isDev = Boolean(dev); // convert string to boolean
+  const isDev = Boolean(dev);
   const {
     showMemberRoleUpdateModal,
     setShowMemberRoleUpdateModal,
@@ -60,17 +60,14 @@ const MemberRoleUpdate = () => {
     setIsUpdating(true);
     let body = {};
     if (archived) {
-      // if user is already archived
       body = {
         archived: false,
       };
     } else if (!reason && !archived) {
-      // if user is not archived and reason is not provided
       body = {
         archived: true,
       };
     } else {
-      // if user is not archived and reason is provided
       body = {
         archived: true,
         reason,
@@ -98,10 +95,10 @@ const MemberRoleUpdate = () => {
     </button>
   );
   const handleValidReason = (id, reason) => {
-    const isEmptyReason = !reason.length; // check for empty string
+    const isEmptyReason = !reason.length;
     const isReasonEmptyOrWhitespace = /^\s*$/.test(reason); // check for empty or multiple whitespaces
-    const isMoreThan99Words = reason.split(' ').length > 99; // check for more than 99 words
-    const isMoreThan50Characters = reason.length <= 25; // check for more than 25 characters
+    const isMoreThan99Words = reason.split(' ').length > 99;
+    const isMoreThan50Characters = reason.length <= 25;
     switch (!archived) {
       case isEmptyReason:
         setvalidateError('Reason cannot be empty!');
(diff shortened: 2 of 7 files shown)


src/components/member-role-update/member-role-update.module.scss
```diff
@@ -9,3 +9,24 @@
...
+  }
+  &__error {
+    color: #e13110;
+    text-align: center;
+    font-size: inherit;
+  }
+  &__success {
+    color: #1e7e34;
+    text-align: center;
+    font-size: inherit;
```
Pratiyushkumar, line 30: please use color variable names instead of color codes directly

Pratiyushkumar (changes requested):

src/components/member-role-update/index.js
```diff
@@ -6,10 +6,14 @@ import Spinner from '@components/UI/spinner';
 import MemberTagAssign from '@components/member-tag-assign';
 import { BASE_API_URL } from '@constants/AppConstants';
 import useFetch from '@custom-hooks/useFetch';
+import { useRouter } from 'next/router';
 import classNames from './member-role-update.module.scss';
 import { memberRoleUpdate } from '../../helper-functions/action-handlers';
 
 const MemberRoleUpdate = () => {
+  const { query } = useRouter() || { query: { dev: false } };
+  const { dev } = query;
```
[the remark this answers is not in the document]
gauravsinhaweb: both ways will function the same, it's the developer's choice

src/components/member-role-update/index.js
```diff
@@ -6,10 +6,14 @@ import Spinner from '@components/UI/spinner';
...
+  const isDev = Boolean(dev); // convert string to boolean
```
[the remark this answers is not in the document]
gauravsinhaweb: tell me why I should use this way

src/components/member-role-update/index.js
```diff
@@ -51,22 +56,31 @@ const MemberRoleUpdate = () => {
     }
   };
 
-  const archiveUnArchiveTheMember = async (id) => {
-    let archiveRole = null;
+  const archiveUnArchiveTheMember = async (id, reason) => {
     setIsUpdating(true);
+    let body = {};
```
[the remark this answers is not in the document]
gauravsinhaweb: because it is supposed to pass as a body in the archiving API, Let me know if you have a better name

src/components/member-role-update/index.js
```diff
@@ -75,30 +89,86 @@ const MemberRoleUpdate = () => {
...
     >
       {archived ? 'Unarchive Member' : 'Archive Member'}
     </button>
   );
 
+  const archiveReasonTextBox = (
+    <div className={classNames.archiveUser}>
+      <p className={classNames.archiveUser__error}>{validateError}</p>
+      <label htmlFor="archiveReason" data-testid="reasonInputLabel">
+        Reason:{' '}
```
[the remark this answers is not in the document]
gauravsinhaweb: updated

src/components/member-role-update/member-role-update.module.scss
gauravsinhaweb: updated

src/components/member-role-update/index.js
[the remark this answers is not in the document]
gauravsinhaweb: You're correct. If we don't need to worry about type coercion and only want to check if the dev property is a string "true," we can perform a conditional check. But what if we encounter values like null, undefined, or even the boolean true,
the code will proceed to the else block. The `|| { query: { dev: false } }` part was added to handle these non-string values,
Let me know if you still want me to remove `|| { query: { dev: false } }`, with a reason

src/components/member-role-update/index.js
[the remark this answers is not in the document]
gauravsinhaweb: I have addressed below

src/components/member-role-update/index.js
```diff
@@ -51,22 +56,31 @@ const MemberRoleUpdate = () => {
...
+        archived: false,
+      };
+    } else if (!reason && !archived) {
+      // if user is not archived and reason is not provided
+      body = {
+        archived: true,
+      };
     } else {
-      archiveRole = true;
+      // if user is not archived and reason is provided
```
sahsisunny, line 73: These comments are important?

src/components/member-role-update/index.js
```diff
@@ -75,30 +89,86 @@ const MemberRoleUpdate = () => {
...
+      default:
+        setvalidateError('');
+    }
+    const isValid =
+      !isEmptyReason &&
+      !isReasonEmptyOrWhitespace &&
+      !isMoreThan99Words &&
+      !isMoreThan50Characters;
+    if (isValid || archived) {
+      archiveUnArchiveTheMember(id, reason); // if valid reason or unarchiving, then archive/unarchive
```
sahsisunny, line 127: Here also. There is no need to put comments for logic you implemented.

gauravsinhaweb: updated

src/components/member-role-update/index.js
gauravsinhaweb: no, updated

.env.development
Pratiyushkumar, line 1: why this change??

Pratiyushkumar (changes requested):

.env.development
gauravsinhaweb: fixed

Pratiyushkumar (approved):

sahsisunny (approved):

iamitprakash (approved):

